PR #197, opened 2026-04-04
[Web] Fix agree/disagree vote button state out of sync across reports

### LINKED ISSUE #196: [Web] Fix agree/disagree vote button state out of sync across reports
# 📝 Issue Title

## Description
The agree/disagree vote buttons on the report panel have two bugs:

1. **Cross-report state leak**: Clicking agree on report A (turns green), then opening report B and clicking agree (turns green), then returning to report A — the agree button reverts to grey despite the user having voted.
2. **Inverted toggle**: Because the frontend initializes vote state as `null` on every mount, it cannot detect whether the user has previously voted (in a prior session). The backend toggles the vote server-side, so clicking agree on an already-agreed report removes the vote — but the frontend shows the button as green (agreed), contradicting the actual state.

## Feature / Task Details

- `userVote` state was local to `ReportPanel`, resetting to `null` on every remount (i.e. every time a different report is selected).
- The toggle logic `setUserVote(prev => prev === type ? null : type)` assumes the frontend knows the prior vote, which it does not after a remount or a fresh page load.
- The backend `/verify` and `/unverify` endpoints are toggle-based (if already agreed → removes agree).

## Deliverables

- Lift vote state to `Home` as a `userVotes` map (`{ [reportId]: 'agree' | 'disagree' | null }`)
- Pass `userVote` and `onVoteChange` as props to `ReportPanel`
- Derive the new vote state from the count delta in the API response (if `agrees` went up → user is now agreed; if down → un-agreed)

## Acceptance Criteria

- [ ] Voting on report A, switching to report B, then returning to A still shows the correct voted state
- [ ] Button turns green immediately on agree and grey on un-agree (no inversion)
- [ ] Vote state is consistent with the server response count
- [ ] All tests pass

## Notes
Fixed in branch `yeldos/web-map-integration`. Root cause: `userVote` was component-local state that reset on every report switch, and the toggle direction was inferred from frontend state rather than the actual server response.

delos0:
# 📝 Description
Fixes #196

Fixes two bugs with the agree/disagree vote buttons on the report panel:

1. **Cross-report state leak** — voting on report A, switching to B, then returning to A reset the button to grey even though the user had voted.
2. **Inverted toggle** — the frontend initialized `userVote` as `null` on every mount. Since the backend toggles server-side, clicking agree on a previously-agreed report removed the vote but the button showed green.

**Root cause:** `userVote` was local state inside `ReportPanel`, resetting on every remount (every report switch). The toggle direction was inferred from frontend state rather than the actual server response.

**Fix:**
- Lifted vote state to `Home` as a `userVotes` map (`{ [reportId]: 'agree' | 'disagree' | null }`)
- Passed `userVote` and `onVoteChange` as props to `ReportPanel`
- Derived new vote state from the count delta in the API response

# 📊 Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation

- [ ] All tests passed

# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/frontend/src/components/ReportPanel.test.jsx
+++ b/frontend/src/components/ReportPanel.test.jsx
@@ -23,12 +23,25 @@ const MOCK_REPORT = {
   longitude: 29.0505,
 }
 
-function renderPanel({ report = MOCK_REPORT, token = null, onClose = vi.fn(), onVoteUpdate = vi.fn() } = {}) {
+function renderPanel({
+  report = MOCK_REPORT,
+  token = null,
+  userVote = null,
+  onClose = vi.fn(),
+  onVoteUpdate = vi.fn(),
+  onVoteChange = vi.fn(),
+} = {}) {
   if (token) localStorage.setItem('token', token)
   return render(
     <MemoryRouter>
       <AuthProvider>
-        <ReportPanel report={report} onClose={onClose} onVoteUpdate={onVoteUpdate} />
+        <ReportPanel
+          report={report}
+          userVote={userVote}
+          onClose={onClose}
+          onVoteUpdate={onVoteUpdate}
+          onVoteChange={onVoteChange}
+        />
       </AuthProvider>
     </MemoryRouter>
   )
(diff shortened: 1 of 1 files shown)


efemantarci (approved):
The state-lifting approach is correct and fixes the cross-report leak cleanly. One note: the count-delta logic for inferring vote state (agrees > prevAgrees ? 'agree' : null) can be replaced with updated.userVote directly once #202 lands.
I have updated the tests to have userVote now so they pass.

PR #357, opened 2026-04-07
[Backend] Return shortest route between wheelchair and ramp-assisted routes

### LINKED ISSUE #350: [Backend] Return shortest route between wheelchair and ramp-assisted routes
# 📝 Merge Wheelchair & Ramp-Assisted Routes — Return the Shorter One

## Description

Currently `RouteService.getRouteOptions()` returns **both** a wheelchair route (Route 3) and a ramp-assisted route (Route 4) as separate options. Instead, we should compute both internally, compare their total distances, and surface only the shorter one to the caller.

Both routes should still be fetched from ORS so the comparison is based on real routing data, not estimates.

## Feature / Task Details

The change is contained in `RouteService.getRouteOptions()` (`backend/src/main/java/com/bounswe2026group1/backend/service/RouteService.java`):

- Continue building **Route 3** (pure wheelchair route with obstacle avoidance) and **Route 4** (ramp-assisted multi-leg route) exactly as today.
- After both routes are constructed, compare `distanceMeters` of Route 3 vs Route 4.
- Add only the shorter of the two to the returned list. Keep its existing `routeLabel` unchanged so the frontend knows which type won.
- If Route 4 cannot be built (no ramp found in the bounding box, or any ORS leg fails), fall back to Route 3 as today.
- If Route 3 fails but Route 4 succeeds, return Route 4.

No changes are needed in the controller, DTOs, or database queries.

## Deliverables

- Updated `RouteService.getRouteOptions()` that picks the shorter of the two accessible routes.
- Unit / integration test covering three scenarios:
  1. Wheelchair route is shorter → it is returned.
  2. Ramp-assisted route is shorter → it is returned.
  3. No ramp available → wheelchair route is returned as fallback.

## Acceptance Criteria

- [ ] `POST /api/routes` never returns **both** a wheelchair route and a ramp-assisted route for the same request.
- [ ] The returned route is the one with the smaller `distanceMeters`.
- [ ] If no ramp exists in the bounding box, the wheelchair route is returned (existing behavior preserved).
- [ ] Existing Route 1 and Route 2 (walking routes) are unaffected.
- [ ] All existing tests pass.
- [ ] New unit tests added for the three comparison scenarios above.

## Notes

**Bounding box for reports / avoid polygons (separate question):**

The avoid-polygon bounding box is built in `ObstacleService.buildAvoidPolygons()` — currently it fetches **all** verified obstacle reports with no spatial pre-filter, which means every obstacle in the database is loaded regardless of how far it is from the requested route. A follow-up improvement would be to add a bounding-box pre-filter (similar to `findRampsInBoundingBoxWithStatuses` in `RampReportRepository`) so only obstacles near the actual route are included. The bounding box could be derived from the route's min/max lat-lon with a small padding (e.g. `AVOID_BUFFER_DEGREES` ≈ 0.01°). This would improve query performance as the obstacle dataset grows, and is tracked separately.

**Relevant files:**
- `backend/src/main/java/com/bounswe2026group1/backend/service/RouteService.java` — main change
- `backend/src/main/java/com/bounswe2026group1/backend/service/ObstacleService.java` — ramp bounding box & avoid polygons (reference only)
- `backend/src/main/java/com/bounswe2026group1/backend/repository/ReportRepository.java` — for future bounding-box pre-filter

## Deadline

efemantarci:
# 📝 Return shortest route between wheelchair and ramp-assisted routes

Fixes  #350 
ramp-assisted routing not using web-created ramp reports. Merges wheelchair and ramp-assisted routes into a single "Ramp-Assisted Route" that picks the shorter path by distance. Also improves error messaging when ramp report creation fails.

# 📊 Type of Change
- [x] Bug fix
- [x] Refactoring
- [ ] New feature
- [ ] Documentation

- [x] All tests passed

# 📸 Screenshots / Recordings
N/A

# ⏱️ Estimated Review Time
- [x] 5-15 min
- [ ] < 5 min
- [ ] > 15 min

### DIFF
--- a/backend/src/main/java/com/bounswe2026group1/backend/service/OverpassService.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/service/OverpassService.java
@@ -3,6 +3,7 @@ package com.bounswe2026group1.backend.service;
 import com.bounswe2026group1.backend.exception.RoutingException;
 import com.bounswe2026group1.backend.model.Location;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Service;
 import org.springframework.web.client.RestClient;
 import tools.jackson.databind.JsonNode;
@@ -48,7 +49,7 @@ public class OverpassService {
 
         List<Location[]> stairEndpoints = parseStairEndpoints(response);
         if (stairEndpoints.isEmpty()) {
-            throw new RoutingException(
+            throw new RoutingException(HttpStatus.BAD_REQUEST,
                     "Ramps can currently only be added to stairs. No stair found within "
                     + SEARCH_RADIUS_METERS + "m of the reported location.");
         }
--- a/backend/src/main/java/com/bounswe2026group1/backend/service/OverpassService.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/service/OverpassService.java
@@ -49,7 +49,8 @@ public class OverpassService {
         List<Location[]> stairEndpoints = parseStairEndpoints(response);
         if (stairEndpoints.isEmpty()) {
             throw new RoutingException(
-                    "No stair found within " + SEARCH_RADIUS_METERS + "m of the reported ramp location");
+                    "Ramps can currently only be added to stairs. No stair found within "
+                    + SEARCH_RADIUS_METERS + "m of the reported location.");
         }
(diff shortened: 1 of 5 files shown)


alaaddinerenn (approved):
Nice fix @efemantarci. That was an urgent issue!

PR #384, opened 2026-04-30
[Backend] Fix data.sql seed: FK-safe comment inserts and disable seed in production

### LINKED ISSUE #383: [Backend] Fix data.sql seed: FK-safe comment inserts and disable seed in production
# 📝 Issue Title
Fix `data.sql` seed: make comment inserts FK-safe and stop running the dev seed in production.

## Description
On 2026-04-30 the production backend container failed to start with this error:

```
Caused by: org.postgresql.util.PSQLException: ERROR: insert or update on table "comments"
  violates foreign key constraint "fkcq6hwtukw527pld91vsq40l0"
  Detail: Key (report_id)=(1) is not present in table "reports".
```

The seed file [`backend/src/main/resources/data.sql`](backend/src/main/resources/data.sql) hardcodes `report_id = 1` (and `2`) in its comment inserts, but report IDs are auto-generated. The `INSERT INTO reports ... WHERE NOT EXISTS (... description = '...')` guard is idempotent on description, so if a report row is ever deleted from prod and re-inserted by the seed on the next boot, it gets a *new* id from the sequence — and the hardcoded `report_id = 1` in the comment insert no longer matches a real row, producing a foreign-key violation that crashes startup.

The file's own first line says **"Seed data for local development"**, yet `application.properties` sets `spring.sql.init.mode=always`, so Spring Boot loads it on every boot — including production. The seed has been quietly running in prod for weeks; it only failed today because data drift produced the orphaned reference.

## Feature / Task Details
Two related fixes:

1. **Make the comment inserts FK-safe.** Look up the parent report by its `description` instead of hardcoding `report_id`. Pattern:
   ```sql
   INSERT INTO comments (content, author_id, report_id, created_at)
   SELECT 'Yes, people also fall from this ramp', 2, r.id, NOW()
   FROM reports r
   WHERE r.description = 'This ramp is too steep for wheelchairs'
     AND NOT EXISTS (
       SELECT 1 FROM comments c
       WHERE c.content = 'Yes, people also fall from this ramp' AND c.report_id = r.id
     );
   ```
   This stays idempotent and degrades to a no-op if the parent report doesn't exist, instead of crashing.

2. **Don't run the dev seed in production.** Make `spring.sql.init.mode` env-driven, default `never`, and opt-in via `SPRING_SQL_INIT_MODE=always` in the local Compose file (`docker-compose.local.yml`). Production `backend/docker-compose.yml` does not set the variable, so the seed never runs there.

## Deliverables
- Updated `data.sql` with description-lookup-based comment inserts.
- `application.properties` switches default `spring.sql.init.mode` to `never` (overridable via `SPRING_SQL_INIT_MODE` env var).
- `docker-compose.local.yml` sets `SPRING_SQL_INIT_MODE: always` for the local backend service.
- Production droplet boots cleanly without the env var (verified by running the new image and confirming `data.sql` is skipped in logs and the app reaches `Started BackendApplication`).

## Acceptance Criteria
- [ ] All comment inserts in `data.sql` resolve `report_id` via a `FROM reports r WHERE r.description = ...` lookup, not a hardcoded id.
- [ ] `application.properties` sets `spring.sql.init.mode=${SPRING_SQL_INIT_MODE:never}`.
- [ ] `docker-compose.local.yml` exports `SPRING_SQL_INIT_MODE=always` to the local backend service so dev seed still works.
- [ ] Local dev (`docker compose -f docker-compose.local.yml up`) still seeds users, reports, and comments in a fresh DB.
- [ ] Prod boot (without the env var) skips the seed and reaches `Started BackendApplication`.
- [ ] Re-running the local seed against an existing DB is idempotent (no duplicate rows, no FK errors).
- [ ] All tests pass.

## Deadline
30.04.2026

efemantarci:
# 📝 Description
Fixes #383

The production backend failed to start on 2026-04-30 with a foreign key violation in the boot seed script:

```
ERROR: insert or update on table "comments" violates foreign key constraint
Key (report_id)=(1) is not present in table "reports".
```

[`backend/src/main/resources/data.sql`](backend/src/main/resources/data.sql) hardcoded `report_id = 1` (and `2`) in its comment inserts, but report ids are auto-generated. The reports seed is idempotent on description, so a deleted-and-re-inserted report comes back with a *new* id from the sequence — at which point the hardcoded comment inserts FK-violate and crash startup. The file's own header says it is "for local development", but `application.properties` had `spring.sql.init.mode=always`, so it ran on every prod boot too.

This PR contains two related fixes:

1. **FK-safe comment inserts.** Every `INSERT INTO comments` now joins the parent report by its `description` and uses `r.id` instead of a hardcoded literal. The insert is still idempotent on `(content, report_id)`, and now also gracefully no-ops if the parent report can't be found rather than crashing.
2. **Default seed off in prod.** `spring.sql.init.mode` is now `${SPRING_SQL_INIT_MODE:never}`. The local Compose file ([docker-compose.local.yml](docker-compose.local.yml)) sets `SPRING_SQL_INIT_MODE: always`, so dev seeding works the same as before. The production Compose file ([backend/docker-compose.yml](backend/docker-compose.yml)) does not set the variable, so `data.sql` simply isn't executed in prod going forward.

# 📊 Type of Change
- [x]  Bug fix
- [ ]  New feature
- [ ]  Refactoring
- [ ]  Documentation

- [x] I have commented my code, especially in hard-to-understand areas (added a header comment above the rewritten comment-insert block explaining why the lookup form is used)
- [ ] I have added/updated tests (n/a — the change is in `data.sql` and a property; tested manually as described below)
- [x] All tests passed

# 📸 Screenshots / Recordings
n/a — config/SQL change. To verify locally:

```bash
# Fresh DB, seed should run and succeed
docker compose -f docker-compose.local.yml down -v
docker compose -f docker-compose.local.yml up -d
docker logs -f mapcess_backend_dev   # expect "Started BackendApplication", no FK errors

# Re-running on an existing DB should be idempotent
docker compose -f docker-compose.local.yml restart backend
docker logs -f mapcess_backend_dev   # still clean, no duplicate-insert errors
```

To verify prod behavior, deploy on the droplet without `SPRING_SQL_INIT_MODE` set; the startup logs should no longer show any `data.sql` script execution.

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min


SimsekMucahit:
I just reviewed and it looks the issue is resolved. Great work on this.

SimsekMucahit (approved):

Issue #365, opened 2026-04-16
[Backend] Accessibility Profiles and Constraint-Based Routing


alaaddinerenn:
# 📝 [Backend] Accessibility Profiles and Constraint-Based Routing

## Description
Implement the backend infrastructure to store user accessibility preferences and apply these constraints when calculating routes or filtering active reports.

## Feature / Task Details
- **User Entity Update:** Add an `accessibilityNeeds` field (e.g., a Set of Enums like `WHEELCHAIR`, `VISUALLY_IMPAIRED`, `NO_STAIRS`, `FLAT_TERRAIN`) to the `RegisteredUser` entity.
- **Routing Engine Constraints:** Update the route calculation logic to append pedestrian accessibility parameters based on the user's profile.
- **Report Filtering Logic:** Ensure that specific report types (e.g., "Broken Elevator", "Construction on Sidewalk") are heavily weighted as hazards for users with the `WHEELCHAIR` or `VISUALLY_IMPAIRED` profiles.

## Deliverables
- Updated database schema for user accessibility preferences.
- Modified routing service that accepts and processes accessibility constraints.
- API endpoints to update user accessibility profiles.

## Acceptance Criteria
- [ ] `GET /api/routes` correctly returns alternative paths when `accessibilityNeeds` are active (e.g., routing around stairs).
- [ ] User preferences are securely persisted in the database.
- [ ] The system accurately correlates specific hazard reports with relevant accessibility profiles.
- [ ] All tests pass
- [ ] Documentation updated (if applicable)

## Deadline
05.05.2026

PR #385, opened 2026-04-30
[Backend] Improve Pathfinding: Multi-Ramp Evaluation, Real Ramp Routing & Accessible Route Fixes

### LINKED ISSUE #375: [Backend] Improve Pathfinding: Multi-Ramp Evaluation, Real Ramp Routing & Accessible Route Fixes
# 📝 Improve Pathfinding: Multi-Ramp Evaluation, Real Ramp Routing & Accessible Route Fixes

## Description
The current routing system delegates graph traversal to OpenRouteService (ORS) and layers accessibility logic on top. Analysis of the codebase revealed several significant issues that reduce the quality and reliability of the ramp-assisted and accessible routes. This issue addresses the most impactful ones for the Final Milestone.

## Feature / Task Details

- **Multi-ramp evaluation**: `ObstacleService.findClosestRampInBoundingBox()` currently picks only the single ramp nearest to the *start point* by Euclidean distance. It should evaluate all candidate ramps and select the one that minimizes total route distance (start→rampEntry + rampExit→end).

- **Real ramp-leg routing**: Leg 2 of the ramp-assisted route (`RouteService.java` ~line 148) is a straight-line segment between ramp entry and exit, which can pass through walls or buildings. It should be replaced with a real ORS `foot-walking` call.

- **Always compute accessible route**: Route 2 (accessible/obstacle-avoiding route) is only computed when Route 1 has verified obstacles (RouteService.java ~line 58). It should always be computed so users always have an accessible alternative.

- **Fix geodesic avoid-polygon buffer**: `AVOID_BUFFER_DEGREES = 0.00009` in `ObstacleService` is only accurate at mid-latitudes. At Istanbul's latitude (~41°N), the buffer is geometrically rectangular. The buffer should use cosine-corrected longitude offset to produce accurate squares.

- **Increase Overpass stair search radius**: `SEARCH_RADIUS_METERS = 5` in `OverpassService` is too strict — ramp report creation fails if the user clicks more than 5m from an OSM stair. Increase to ~15m for a better UX.

- **Wire up `mode` field**: `RouteRequest.mode` is always sent as `null` from the frontend and is ignored by `RouteService`. Hooking this up would allow users (and future accessibility profile logic) to request specific route types.

## Deliverables

- Updated `ObstacleService.findClosestRampInBoundingBox()` to evaluate all ramps and return the optimal one by total route length
- Updated `RouteService` ramp-assisted route to use an ORS call for the ramp traversal leg instead of a straight-line segment
- Updated `RouteService` to always compute the accessible route regardless of obstacle presence on the fastest route
- Corrected avoid-polygon buffer calculation in `ObstacleService` using latitude-aware geodesic offsets
- Increased Overpass search radius from 5m to 15m in `OverpassService`

## Acceptance Criteria

- [ ] Ramp-assisted route selects the ramp that minimizes total journey distance, not just nearest to start
- [ ] The ramp traversal segment is a real routed path (no straight-line shortcuts through buildings)
- [ ] An accessible route option is always returned regardless of obstacle presence on the fastest route
- [ ] Avoid polygons are geometrically square (not rectangular) at Istanbul's latitude
- [ ] Ramp report creation succeeds when user clicks within 15m of an OSM stair
- [ ] All existing routing tests pass

## Deadline

2026-05-14 (Milestone 3: Final Milestone)

## Notes

Related issues: #365 (Accessibility Profiles and Constraint-Based Routing). The `mode` field wiring may overlap with that issue and should be coordinated.

efemantarci:
# 📝 Description
Fixes #375

Improves the quality and reliability of ramp-assisted and accessible routing. Five backend changes plus tests:

1. **Multi-ramp evaluation by total route distance.** [`ObstacleService.findClosestRampInBoundingBox()`](backend/src/main/java/com/bounswe2026group1/backend/service/ObstacleService.java) now evaluates every candidate ramp and picks the one minimizing `haversine(start → ramp entry) + haversine(ramp exit → end)`, rather than only the ramp whose entry is nearest to the start. The previous heuristic could pick a ramp adjacent to the start but pointing entirely the wrong way.
2. **Real ramp-leg routing.** [`RouteService`](backend/src/main/java/com/bounswe2026group1/backend/service/RouteService.java) leg 2 (the traversal of the ramp itself) used to be a synthetic straight line generated by a `straightLineLeg(...)` helper, which could cut through walls. It now uses an ORS `WALKING` call so the ramp leg follows real walkable geometry. The helper is removed.
3. **Accessible route is always computed when avoid polygons exist.** The guard at [RouteService.java:54](backend/src/main/java/com/bounswe2026group1/backend/service/RouteService.java#L54) was `if (hasObstacles && avoidPolygons != null)`, so users only saw an accessible alternative if the fastest route happened to cross a reported obstacle. It's now `if (avoidPolygons != null)`, so users always have an accessible option when any obstacles are reported in the area. Cost: at most one extra ORS call per request.
4. **Cosine-corrected avoid-polygon buffer.** [`ObstacleService.squareRingAround`](backend/src/main/java/com/bounswe2026group1/backend/service/ObstacleService.java) now applies `dLon = dLat / cos(lat)` so avoid polygons are geographically square at Istanbul's latitude (~41°N), not rectangular as they were with a single fixed-degree buffer.
5. **Overpass stair search radius 5 m → 15 m.** [`OverpassService`](backend/src/main/java/com/bounswe2026group1/backend/service/OverpassService.java) was rejecting ramp reports when the user clicked more than 5 m from the nearest OSM stair, which was much too tight in practice.

The 6th item from #375 (wiring up `RouteRequest.mode`) is intentionally **out of scope** for this PR — the issue notes that work overlaps with #365 (Accessibility Profiles and Constraint-Based Routing) and is best done there.

## Tests added
- [`ObstacleServiceTest`](backend/src/test/java/com/bounswe2026group1/backend/service/ObstacleServiceTest.java)
  - `findClosestRamp_returnsNull_whenNoCandidates`
  - `findClosestRamp_returnsSingleRamp_whenOnlyOneCandidate`
  - `findClosestRamp_picksRampWithLowerTotalRouteEstimate_notJustNearestToStart` — the key correctness test, with rampA close to start but useless for the destination, and rampB the better choice by total distance
  - `findClosestRamp_skipsRampsWithNullEntryOrExit`
  - `buildAvoidPolygons_returnsNull_whenNoObstacles`
- [`RouteServiceTest`](backend/src/test/java/com/bounswe2026group1/backend/service/RouteServiceTest.java)
  - `accessibleRoute_isComputed_whenAvoidPolygonsExist_evenIfFastestRouteHasNoObstacles` — verified to fail against the old `hasObstacles && ...` guard and pass against the new one

# 📊 Type of Change
- [x]  Bug fix
- [x]  New feature
- [ ]  Refactoring
- [ ]  Documentation

- [x] All tests passed (`./mvnw -Dtest='ObstacleServiceTest,RouteServiceTest' test` → 6/6 green)

# 📸 Screenshots / Recordings
n/a — backend logic change. Behavior to verify manually:
- Ramp-assisted route prefers a ramp with a useful exit direction over a ramp closest to the start (you can construct a scenario by adding ramp reports across campus).
- An "Accessible Route" option appears whenever any obstacle exists in the area, even when the fastest route on screen has no obstacles on it.
- Reporting a ramp 5–15 m from the nearest OSM stair now succeeds instead of returning 400.

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min


backend/src/main/java/com/bounswe2026group1/backend/service/RouteService.java
```diff
@@ -99,7 +96,7 @@ public List<RouteResponse> getRouteOptions(RouteRequest request) {
             Location rampExit  = distAToStart <= distBToStart ? rampB : rampA;
 
             RoutingDirectionsResult leg1 = fetchOrNull(start, rampEntry, TravelMode.WHEELCHAIR, avoidPolygons);
-            RoutingDirectionsResult leg2 = straightLineLeg(rampEntry, rampExit);
+            RoutingDirectionsResult leg2 = fetchOrNull(rampEntry, rampExit, TravelMode.WALKING, null);
             RoutingDirectionsResult leg3 = fetchOrNull(rampExit, end, TravelMode.WHEELCHAIR, avoidPolygons);
 
             if (leg1 != null && leg3 != null) {
```
alaaddinerenn: leg2 can now return null — please add a null check for it here.

alaaddinerenn (changes requested):
Could you apply the changes that I requested?

backend/src/main/java/com/bounswe2026group1/backend/service/RouteService.java
efemantarci: I have added the null check updated tests

alaaddinerenn (approved):
LGTM! Thanks for making the requested changes — everything looks good now. Approving.

PR #388, opened 2026-04-30
[Backend] Implement Update Report Endpoint

### LINKED ISSUE #175: [Backend] Implement Report Edit and Delete Endpoints
# 📝 Issue Title

## Description
Implement the backend REST endpoints that allow users to update the details of an existing accessibility report or permanently delete it. This feature requires strict authorization checks to ensure that users can only modify or delete content they own, while Admins maintain global override privileges.

## Feature / Task Details

- Implement a PUT /api/reports/{id} endpoint to allow updating a report's description, tag, or status. (Determine if modifying coordinates latitude/longitude after creation is allowed, or if they should create a new report).
- Implement a DELETE /api/reports/{id} endpoint to permanently remove a report from the database.
- Security & Authorization: Validate the JWT token of the requester against the user_id of the report. Deny the request if the user is not the original creator (unless they have the ADMIN role).
- Cascading Deletion: Update the ReportService to ensure that when a report is deleted, a call is made to the S3MediaService to physically delete any associated images/videos from the AWS bucket so we do not leave orphaned files behind.

## Deliverables

- PUT and DELETE endpoints in the ReportController.
- Update and delete logic in the ReportService, including S3 cleanup.
- Security checks verifying ownership before executing modifications.
- Unit and integration tests (JUnit + Mockito + MockMvc) verifying successful operations, ownership rejection (403), and missing reports (404).

## Acceptance Criteria

- [ ] PUT returns 200 OK with the updated report object upon success.
- [ ] DELETE returns 204 No Content (or 200 OK) upon successful deletion, and associated S3 files are removed.
- [ ] Returns 403 Forbidden if a user attempts to edit or delete a report they did not create.
- [ ] Returns 404 Not Found (throwing NoSuchElementException) if the {id} does not exist.
- [ ] All tests pass

## Deadline
30.04.2026

alaaddinerenn:
# 📝 Description
Implements `PUT /api/reports/{id}` endpoint allowing any authenticated user to update a report's fields and remove attached media. Also fixes Spring Security misconfiguration that caused error responses to return 401. Fixes #175 

# 📊 Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

- [x] All tests passed

# 📸 Screenshots / Recordings
N/A

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/src/main/java/com/bounswe2026group1/backend/controller/ReportController.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/controller/ReportController.java
@@ -50,11 +50,9 @@ public class ReportController {
     }
 
     @DeleteMapping("/{id}")
-    public ResponseEntity<Void> delete(@PathVariable Long id) {
-        if (reportService.delete(id)) {
-            return ResponseEntity.noContent().build();
-        }
-        return ResponseEntity.notFound().build();
+    @ResponseStatus(HttpStatus.NO_CONTENT)
+    public void delete(@PathVariable Long id, @AuthenticationPrincipal String email) {
+        reportService.delete(id, email);
     }
 
     @PostMapping("/{id}/verify")
--- a/backend/src/main/java/com/bounswe2026group1/backend/controller/ReportController.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/controller/ReportController.java
@@ -2,6 +2,7 @@ package com.bounswe2026group1.backend.controller;
 
 import com.bounswe2026group1.backend.dto.CreateReportRequest;
 import com.bounswe2026group1.backend.dto.ReportResponse;
+import com.bounswe2026group1.backend.dto.UpdateReportRequest;
 import com.bounswe2026group1.backend.service.ReportService;
 import lombok.RequiredArgsConstructor;
 import org.springframework.http.HttpStatus;
@@ -43,10 +44,9 @@ public class ReportController {
     }
 
     @PutMapping("/{id}")
-    public ResponseEntity<ReportResponse> update(@PathVariable Long id, @RequestBody CreateReportRequest request) {
-        return reportService.update(id, request)
-                .map(ResponseEntity::ok)
-                .orElseGet(() -> ResponseEntity.notFound().build());
+    public ResponseEntity<ReportResponse> update(@PathVariable Long id,
+                                                  @RequestBody UpdateReportRequest request) {
+        return ResponseEntity.ok(reportService.update(id, request));
     }
 
     @DeleteMapping("/{id}")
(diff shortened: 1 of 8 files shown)


backend/src/main/java/com/bounswe2026group1/backend/controller/ReportController.java
```diff
@@ -43,18 +44,15 @@ public ResponseEntity<ReportResponse> create(@RequestBody CreateReportRequest re
     }
 
     @PutMapping("/{id}")
-    public ResponseEntity<ReportResponse> update(@PathVariable Long id, @RequestBody CreateReportRequest request) {
-        return reportService.update(id, request)
-                .map(ResponseEntity::ok)
-                .orElseGet(() -> ResponseEntity.notFound().build());
+    public ResponseEntity<ReportResponse> update(@PathVariable Long id,
```
efemantarci, line 47: Delete checks the email so that only the owner of the report can delete it. However, update can be done by anyone. Maybe it should be changed.

backend/src/main/java/com/bounswe2026group1/backend/service/ReportService.java
```diff
@@ -98,23 +102,48 @@ public ReportResponse create(CreateReportRequest request) {
...
-        });
+    @Transactional
+    public ReportResponse update(Long id, UpdateReportRequest request) {
+        Report report = reportRepository.findById(id)
+                .orElseThrow(() -> new ResponseStatusException(HttpStatus.NOT_FOUND, "Report not found with id: " + id));
+
+        report.setDescription(request.getDescription());
+        report.setTag(request.getTag());
+        report.getLocation().setLatitude(request.getLatitude());
+        report.getLocation().setLongitude(request.getLongitude());
```
efemantarci: ReportRequest can have null variables. So if user did not change the tag, tag will be set to null after update. We should set only if it is not null, or make the request @NotNull

efemantarci (changes requested):

backend/src/main/java/com/bounswe2026group1/backend/controller/ReportController.java
alaaddinerenn: This is intended — the update report is meant to be accessible by all users.

backend/src/main/java/com/bounswe2026group1/backend/service/ReportService.java
alaaddinerenn: Nice point, I should add a null check.

backend/src/main/java/com/bounswe2026group1/backend/controller/ReportController.java
efemantarci: Ok nice. Frontend can handle which parts can be changed by others.

backend/src/main/java/com/bounswe2026group1/backend/service/ReportService.java
efemantarci: Ok I saw the new null checks. Thank you

efemantarci (approved):
Nice. We should warn frontend about edit restrictions on other though.

PR #389, opened 2026-04-30
[Backend] User Profile

### LINKED ISSUE #302: [Backend] User Profile
# [Backend] User Profile

Parent issue: #TBD

## Description

Implement backend support for the user profile feature in MapCess. The Spring Boot backend must expose REST endpoints to fetch and update user profile data, and handle avatar image uploads, so that profile information persists across sessions and devices.

## Feature / Task Details

- Extend the user model / table to include: `displayName`, `bio`, `avatarUrl`
- Implement REST endpoints:
  - `GET /api/users/{id}/profile` — retrieve full profile for the authenticated user (or any user for public view)
  - `PUT /api/users/{id}/profile` — update display name, bio for the authenticated user
  - `POST /api/users/{id}/profile/avatar` — upload and store a new avatar image; return the hosted URL
- Reuse the existing media upload service (see PR #109) for avatar storage
- Include contribution stats in the GET response: count of accessibility reports submitted, routes planned
- Secure endpoints: only authenticated users may update their own profile
- Return profile data (including `avatarUrl`) as part of the login/me response to minimise extra round-trips
- Write unit and integration tests for all new endpoints

## Deliverables

- Database migration for new profile fields (`displayName`, `bio`, `avatarUrl`)
- `GET /api/users/{id}/profile` and `PUT /api/users/{id}/profile` endpoints
- `POST /api/users/{id}/profile/avatar` endpoint using media upload service
- Profile data included in the login/me response payload
- Unit and integration tests

## Acceptance Criteria

- [ ] `GET /api/users/{id}/profile` returns full profile including contribution stats
- [ ] `PUT /api/users/{id}/profile` updates and persists display name and bio
- [ ] `POST /api/users/{id}/profile/avatar` accepts an image, stores it, and returns the URL
- [ ] Endpoints are secured — users cannot modify other users' profiles
- [ ] Profile data is returned in the login/me response
- [ ] Invalid input is rejected with appropriate 400 errors
- [ ] Unit and integration tests written and passing
- [ ] API documented (Swagger/OpenAPI updated)

## Deadline

01.05.2026

## Notes

Avatar storage should reuse the media upload service merged in PR #109. Agree on the profile API contract (field names, avatar upload format) with the web and mobile teams before implementation.

efemantarci:
# 📝 [Backend] User Profile
Fixes #302

Adds the backend API for the User Profile feature.

- New endpoints: `GET /api/users/me`, `GET/PUT /api/users/{id}/profile`, `POST/DELETE /api/users/{id}/profile/avatar`.
- `LoginResponse` embeds a `UserProfileDTO` so clients have the full profile in the login response.
- `Route` gets a `createdBy` owner; `RouteController` records a row per `/api/routes` call from authenticated users so the `routesPlanned` contribution stat is real.
- New `GET /api/routes/user/{userId}` returns the user's routes (mirrors `GET /api/reports/user/{userId}`), so the frontend can render a user's route history. Reports are already covered by `ReportController`.
- `S3MediaService.deleteFile(url)` is reused by `DELETE /profile/avatar` to remove the underlying S3 object alongside the URL.
- Fixes a pre-existing config bug: `/error` was missing from `SecurityConfig` permitAll, so `@Valid` 400s got masked as 401 in real Tomcat (slice tests with MockMvc didn't catch it).

Drive-by: existing `GET /api/users` and `GET /api/users/{id}` now return `UserProfileDTO` instead of the raw `RegisteredUser`, closing a password-hash leak.

**Out of scope (separate tickets):** avatar replace-cleanup (orphans on re-upload), public-vs-private profile flag, username/handle field, pagination on `GET /api/routes/user/{userId}`, backfilling `routes.created_by` for existing anonymous routes.

# 📊 Type of Change
- [ ]  Bug fix
- [x]  New feature
- [ ]  Refactoring
- [ ]  Documentation

- [x] All tests passed

# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [ ] 5-15 min
- [x] > 15 min

### DIFF
--- a/backend/src/main/java/com/bounswe2026group1/backend/controller/RegisteredUserController.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/controller/RegisteredUserController.java
@@ -32,9 +32,7 @@ public class RegisteredUserController {
 
     @GetMapping
     public List<UserProfileDTO> getAll() {
-        return registeredUserService.getAll().stream()
-                .map(u -> registeredUserService.getProfileById(u.getId()))
-                .toList();
+        return registeredUserService.getAllProfiles();
     }
 
     @GetMapping("/{id}")
--- a/backend/src/main/java/com/bounswe2026group1/backend/controller/RegisteredUserController.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/controller/RegisteredUserController.java
@@ -11,6 +11,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.AccessDeniedException;
+import org.springframework.security.authentication.AnonymousAuthenticationToken;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.web.bind.annotation.*;
@@ -68,15 +69,6 @@ public class RegisteredUserController {
         }
         try {
             return ResponseEntity.ok(registeredUserService.getProfileByEmail(email));
-        } catch (NoSuchElementException e) {
-            return ResponseEntity.status(HttpStatus.NOT_FOUND).body(e.getMessage());
-        }
-    }
-
-    @GetMapping("/{id}/profile")
-    public ResponseEntity<?> getProfile(@PathVariable Long id) {
-        try {
-            return ResponseEntity.ok(registeredUserService.getProfileById(id));
         } catch (NoSuchElementException e) {
             return ResponseEntity.status(HttpStatus.NOT_FOUND).body(e.getMessage());
         }
@@ -156,7 +148,10 @@ public class RegisteredUserController {
 
     private String currentUserEmailOrNull() {
         Authentication auth = SecurityContextHolder.getContext().getAuthentication();
-        if (auth == null || !auth.isAuthenticated() || auth.getName() == null) return null;
+        // Spring's anonymous token reports isAuthenticated() == true with name "anonymousUser".
+        if (auth == null || auth instanceof AnonymousAuthenticationToken || !auth.isAuthenticated() || auth.getName() == null) {
+            return null;
+        }
         return auth.getName();
     }
 }
--- a/backend/src/main/java/com/bounswe2026group1/backend/controller/RegisteredUserController.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/controller/RegisteredUserController.java
@@ -7,6 +7,7 @@ import com.bounswe2026group1.backend.service.RegisteredUserService;
 import com.bounswe2026group1.backend.service.S3MediaService;
 import jakarta.validation.Valid;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.AccessDeniedException;
@@ -19,6 +20,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 
+@Slf4j
 @RestController
 @RequestMapping("/api/users")
 @RequiredArgsConstructor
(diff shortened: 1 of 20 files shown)


backend/src/main/java/com/bounswe2026group1/backend/dto/UserProfileDTO.java
```diff
@@ -0,0 +1,30 @@
...
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class UserProfileDTO {
+
+    private Long id;
+    private String name;
+    private String email;
```
alaaddinerenn, line 16: GET /{id}/profile requires authentication, but any logged-in user can still look up another user's email address by iterating over IDs. Email is essentially PII — consider removing it from this DTO or restricting it to the /me response only, where a user is reading their own data.

backend/src/main/java/com/bounswe2026group1/backend/controller/RegisteredUserController.java
```diff
@@ -1,30 +1,48 @@
...
     @GetMapping
-    public List<RegisteredUser> getAll() {
-        return registeredUserService.getAll();
+    public List<UserProfileDTO> getAll() {
+        return registeredUserService.getAll().stream()
+                .map(u -> registeredUserService.getProfileById(u.getId()))
+                .toList();
     }
 
     @GetMapping("/{id}")
```
alaaddinerenn, line 38: GET /{id} and GET /{id}/profile are identical — both return UserProfileDTO for the same user. Having two endpoints with the same behavior creates ambiguity and doubles the maintenance surface. Please remove one and document which is the canonical URL.

backend/src/main/java/com/bounswe2026group1/backend/controller/RegisteredUserController.java
```diff
@@ -39,4 +57,106 @@ public ResponseEntity<Void> delete(@PathVariable Long id) {
...
+        }
+        UserProfileDTO me = registeredUserService.getProfileByEmail(email);
+        if (!me.getId().equals(pathId)) {
+            throw new AccessDeniedException("You can only modify your own profile.");
+        }
+    }
+
+    private String currentUserEmailOrNull() {
+        Authentication auth = SecurityContextHolder.getContext().getAuthentication();
+        if (auth == null || !auth.isAuthenticated() || auth.getName() == null) return null;
```
alaaddinerenn: auth.isAuthenticated() returns true for Spring Security's anonymous authentication token — so this check won't catch unauthenticated requests. auth.getName() will return "anonymousUser" instead of null, causing getProfileByEmail("anonymousUser") to throw downstream. Should be:


if (auth == null || auth instanceof AnonymousAuthenticationToken || !auth.isAuthenticated()) return null;

alaaddinerenn (changes requested):
Good work. Three small things I've commented on inline — worth addressing before merge.

backend/src/main/java/com/bounswe2026group1/backend/dto/UserProfileDTO.java
efemantarci: Now DTO takes includeEmail flag. So we can still send email info from /me or PUT /{id}/profile. But other calls such as GET/{id} does not show email

backend/src/main/java/com/bounswe2026group1/backend/controller/RegisteredUserController.java
efemantarci: You are right. It rejects anonymous tokens now. Thank you for the comment

backend/src/main/java/com/bounswe2026group1/backend/controller/RegisteredUserController.java
efemantarci: You are right. I am keeping GET /{id} since mobile uses it. Other will be removed.

alaaddinerenn (approved):
Thanks for addressing the feedback and making the requested changes — I appreciate you taking the time to update this. The code looks clean and everything is in order.

Issue #316, opened 2026-04-06
[Backend] Follow Updates: Backend Implementation & Frontend Integration


delos0:
# 📝 Follow Updates — Backend & Frontend Integration

## Description
The Follow Updates button in ReportPanel (#308) was wired with a UI toggle, but has no backend persistence or initial state hydration. This issue tracks the full implementation of the follow/unfollow feature end-to-end.

## Feature / Task Details

- Implement a backend endpoint to follow/unfollow a report (e.g. `POST /reports/{id}/follow`, `DELETE /reports/{id}/follow`)
- Return the current follow status when fetching a report (or via a dedicated `GET /reports/{id}/follow` endpoint)
- Wire the frontend Follow Updates button in `ReportPanel.jsx` to call the API on click
- Initialize the button state from the backend on panel open (so returning users see the correct state)
- Handle loading and error states in the UI

## Deliverables

- Backend REST endpoints for follow/unfollow
- Frontend API service function(s) for follow/unfollow
- `ReportPanel.jsx` updated to persist and hydrate follow state via API

## Acceptance Criteria

- [ ] Authenticated users can follow a report and the state persists after page refresh
- [ ] Authenticated users can unfollow a report
- [ ] Panel opens showing the correct follow state for the current user
- [ ] Guest users are redirected to login (already implemented in #308)
- [ ] Loading state is shown while the API call is in progress
- [ ] Errors are surfaced to the user if the API call fails
- [ ] All tests pass
- [ ] Documentation updated (if applicable)

## Deadline

01.05.2026

## Notes

- Related PR: #308 (UI toggle — merged without backend)
- Related issue: #305

efemantarci:
Closed by pr #391

PR #391, opened 2026-05-01
[Backend] Notification Entity, Repository, and In-App Triggers

### LINKED ISSUE #258: [Backend] Notification Entity, Repository, and In-App Triggers
# 📝 [Backend] Notification Entity, Repository, and In-App Triggers

## Description
Implement the core backend logic to generate and store notifications. This covers the persistence of alerts and the automated triggers within the system's business logic.

## Feature / Task Details
- **Entity Creation:** Create a `Notification` JPA entity with fields: `id`, `recipientId`, `message`, `type` (STATUS_CHANGE, NEW_COMMENT, NAV_ALERT), `isRead` (boolean), and `createdAt`.
- **Status Change Trigger:** Implement a service listener that creates a notification when a report status changes to "Verified" or "Rejected".
- **Comment Trigger:** Create a notification for the report author when another user submits a comment on their report.
- **Notification API:** - `GET /api/notifications`: Retrieve all notifications for the authenticated user.
  - `PATCH /api/notifications/{id}/read`: Mark a specific notification as read.

## Deliverables
- `Notification` JPA Entity and Repository.
- Automated service triggers for status changes and comments.
- REST endpoints for notification management.
- Unit tests for the notification service logic.

## Acceptance Criteria
- [ ] Notifications are successfully persisted in the database upon triggers.
- [ ] Users can only fetch their own notifications.
- [ ] Marking a notification as read updates the `isRead` flag in the DB.
- [ ] All tests pass

## Deadline
20.04.2026

## Notes
Ensure that creating a notification does not block the main transaction (consider using `@Async` for notification generation).

efemantarci:
# 📝 Description
Fixes #258
Refs #316 (backend half — frontend wiring of `ReportPanel.jsx` is still pending)

Adds the backend Notification feature plus per-user SSE delivery and an explicit "Follow Updates" subscription model so notifications fan out to author, commenters, voters, and followers.

**Persistence**
- `Notification` entity (recipient, message, type, isRead, createdAt, relatedEntityId).
- `NotificationType` enum: `STATUS_CHANGE`, `NEW_COMMENT`, `NAV_ALERT`. `NAV_ALERT` is reserved — no producer wired yet.
- `ReportSubscription` entity with unique (user, report) so following twice is a no-op. The entity / table keep the "Subscription" naming for DB stability under `ddl-auto=update`; the HTTP layer uses "follow".

**Triggers**
- `ReportService.verifyReport` / `unverifyReport` fire `notifyStatusChange` only when the status actually transitions to `VERIFIED` or `REJECTED`. The voting user is passed as the actor.
- `CommentService.create` fires `notifyNewComment` after persistence.
- Both call paths use the existing post-commit broadcast pattern, so the DB row is durable before the SSE push fires.

**Audience fan-out**
- `NotificationService` builds `{ author ∪ commenters ∪ voters ∪ followers } − actor` (deduped) and persists one notification per recipient. The author gets the personalized "Your report" wording; everyone else gets "report you follow".

**REST API** (all authenticated)
- `GET /api/notifications` — current user's list, newest first.
- `PATCH /api/notifications/{id}/read` — flips `isRead`. 403 if not the recipient.
- `POST /api/reports/{id}/follow` — idempotent.
- `DELETE /api/reports/{id}/follow` — idempotent.
- `GET /api/reports/{id}/follow/me` — `{ following: bool }`.

`SecurityConfig` gains an explicit `authenticated()` rule for `/api/reports/*/follow/**` that runs before the existing permit-all `GET /api/reports/**`, so the more specific match wins.

**SSE (additional, not in the issue spec)**
- `NotificationSseService` is a new per-user emitter map (separate from the broadcast `PublicSseService`); only the recipient receives their events. Same connection-limit and cleanup pattern as `PublicSseService`, configurable via `app.sse.notifications.max-connections` and `app.sse.notifications.max-connections-per-user`.
- `GET /api/sse/notifications/subscribe` returns the stream. Browser `EventSource` cannot set custom headers, so `JwtAuthFilter` now also accepts a `?token=` query parameter — restricted to `/api/sse/notifications/**` so other endpoints stay header-only.
- `AsyncConfig` adds an `@EnableAsync` + dedicated `notificationExecutor`; SSE pushes run on it so client I/O latency does not extend the trigger transaction.

**Other notes**
- Status-change message lowercases via `Locale.ROOT` so `VERIFIED` does not become `verıfıed` on Turkish JVMs.
- No mobile or web wiring — the frontend team owns the "Follow Updates" button against the new endpoints (#316 covers `ReportPanel.jsx`; mobile has the same no-op button at `report_detail_screen.dart:1294-1329` and will likely need its own ticket).

# 📊 Type of Change
- [x]  New feature

- [x] All tests passed (99 passed, 1 skipped — pre-existing skip)

# 📸 Screenshots / Recordings
Manually verified end-to-end against the local Docker stack:
- `POST /api/comments` from Bob fans out a `NEW_COMMENT` notification to Alice (report author, "Your report" wording) and Carol (follower, "you follow" wording). Bob himself ends up with `[]`.
- `PATCH /api/notifications/{id}/read` flips `read` from false to true.
- `GET /api/sse/notifications/subscribe?token=<jwt>` opens a stream that emits `event:connected` immediately and a fresh `event:notification` payload within ~1s of the next comment.

# ⏱️ Estimated Review Time
- [x] > 15 min


backend/src/main/java/com/bounswe2026group1/backend/config/JwtAuthFilter.java
```diff
@@ -31,15 +31,22 @@ protected void doFilterInternal(HttpServletRequest request,
...
+            jwt = authHeader.substring(7);
+        } else if (isSseRequest(request)) {
+            String tokenParam = request.getParameter("token");
+            if (tokenParam == null || tokenParam.isBlank()) {
+                filterChain.doFilter(request, response);
+                return;
+            }
+            jwt = tokenParam;
+        } else {
             filterChain.doFilter(request, response);
```
alaaddinerenn: Avoid putting long-lived JWT tokens in the URL query string**
`JwtAuthFilter` now reads `?token=` for SSE endpoints. Long-lived JWTs in URLs
are exposed in server access logs, reverse-proxy logs, and browser history
(OWASP A07). Please either (a) add a log-scrubbing filter that redacts the
`token` parameter, or (b) introduce a short-lived one-time SSE token endpoint
(`POST /api/sse/token`) so the durable JWT never appears in a URL.

backend/src/main/java/com/bounswe2026group1/backend/service/NotificationService.java
```diff
@@ -0,0 +1,193 @@
...
+        Long reportId = report.getReportId();
+        Long authorId = report.getCreatedBy().getId();
+
+        for (Long recipientId : resolveAudience(reportId, authorId, actorUserId)) {
+            String message = buildStatusMessage(recipientId, authorId, reportId, statusLower);
+            RegisteredUser recipient = loadRecipient(recipientId);
+            if (recipient == null) continue;
+            create(recipient, message, NotificationType.STATUS_CHANGE, reportId);
+        }
+    }
```
alaaddinerenn, line 96: Fan-out loop issues N+1 `findById` queries `notifyStatusChange` and `notifyNewComment` call `loadRecipient(id)` for every member of the resolved audience. On a high-traffic report this becomes N sequential DB round-trips. Replace the per-recipient `findById` calls with a single `registeredUserRepository.findAllById(audienceIds)` before the loop.

backend/src/main/java/com/bounswe2026group1/backend/model/Notification.java
```diff
@@ -0,0 +1,49 @@
...
+    private NotificationType type;
+
+    @Column(name = "is_read", nullable = false)
+    private boolean isRead = false;
+
+    @Column(name = "related_entity_id")
+    private Long relatedEntityId;
+
+    @Column(name = "created_at", nullable = false, updatable = false)
+    private LocalDateTime createdAt;
```
alaaddinerenn: Use `LocalDateTime.now(ZoneOffset.UTC)` in `@PrePersist` and `NotificationSseEvent.created()`. `LocalDateTime.now()` depends on the JVM default timezone. Be explicit.

backend/src/main/java/com/bounswe2026group1/backend/dto/NotificationSseEvent.java
```diff
@@ -0,0 +1,39 @@
...
+    public static NotificationSseEvent created(
+            Long notificationId,
+            Long recipientId,
+            NotificationType type,
+            String message,
+            Long relatedEntityId,
+            LocalDateTime createdAt
+    ) {
+        Instant ts = createdAt == null ? Instant.now() : createdAt.toInstant(ZoneOffset.UTC);
+        return new NotificationSseEvent(
```
alaaddinerenn: Also note that `NotificationSseEvent` uses `Instant` while `NotificationResponse` uses `LocalDateTime`; a client consuming both channels will see the same notification with two different timestamp representations — align them.

alaaddinerenn (changes requested):
I would appreciate it if you could go ahead and implement the required changes. @efemantarci

backend/src/main/java/com/bounswe2026group1/backend/config/JwtAuthFilter.java
efemantarci: Done. Now there is 60s token endpoint to subscribe and it is stored as a cookie. So it is safe.
I added POST /api/sse/token endpoint for this and opened an issue for the frontend

backend/src/main/java/com/bounswe2026group1/backend/service/NotificationService.java
efemantarci: This is batched now. Tests are updated too.

backend/src/main/java/com/bounswe2026group1/backend/model/Notification.java
efemantarci: It uses `Instant `time now

backend/src/main/java/com/bounswe2026group1/backend/dto/NotificationSseEvent.java
efemantarci: It uses `Instant `time now

alaaddinerenn (approved):
Great work! Really appreciate you going through the feedback and making all the changes. This is looking solid now — happy to approve.

PR #394, opened 2026-05-01
[Backend] Use Instant Instead Of LocalDateTime For API Timestamps

### LINKED ISSUE #221: [Backend] Use Instant Instead Of LocalDateTime For API Timestamps
# 📝 Use Instant Instead Of LocalDateTime For API Timestamps

## Description

The Backend Currently Uses `java.time.LocalDateTime` For Timestamp Fields Exposed Via The API. Jackson Serializes `LocalDateTime` As ISO-8601 **Without A Timezone Suffix** (e.g. `"2026-04-05T14:30:00"`), Which Forces Every Client To Guess Whether The Value Is UTC Or Local Time. This Caused A "Time Ago" Display Bug On Mobile Where Timestamps Appeared Offset By The Device's Timezone.

Mobile Shipped A Workaround In #217 By Appending `Z` To The String On The Dart Side, But This Only Works Because The Production Container Happens To Run In UTC — It's Correct By Accident, And The Same Bug Will Resurface For Any Future Web Client Or Any Non-UTC Backend Environment.

The Proper Fix Is To Switch The Affected Fields To `java.time.Instant`, Which Jackson Serializes As `"2026-04-05T14:30:00Z"` Automatically. Every Client Then Parses The Timestamp Correctly Without Per-Client Workarounds.

## Feature / Task Details

- Replace `LocalDateTime` With `Instant` In The Following Fields:
  - `model/Report.java` → `publishDate` (And The `@PrePersist` Initializer)
  - `model/Comment.java` → `createdAt` (And Its Initializer)
  - `dto/ReportResponse.java` → `publishDate`
  - `dto/RampReportResponse.java` → `publishDate`
- Update Affected Getters/Setters, Mappers, And Any Service-Layer Code That Reads/Writes These Fields
- Verify Postgres Column Types: If Existing Columns Are `TIMESTAMP` (Zoneless), Add A Schema Migration To `TIMESTAMPTZ USING ... AT TIME ZONE 'UTC'`
- Optionally Add `spring.jackson.serialization.write-dates-as-timestamps=false` To `application.properties` (Defensive)
- Verify Swagger/OpenAPI Still Renders Timestamp Fields Correctly

## Deliverables

- Updated Entity And DTO Classes Using `Instant`
- DB Migration Script (If Column Types Require It)
- Verified API Response Shows `Z`-Suffixed Timestamps (e.g. Via `curl` On A Report Endpoint)

## Acceptance Criteria

- [ ] All Timestamp Fields In `Report`, `Comment`, `ReportResponse`, `RampReportResponse` Use `Instant`
- [ ] API JSON Responses Emit Timestamps With An Explicit `Z` Suffix
- [ ] Postgres Columns Are `TIMESTAMPTZ` (Verified Via `\d reports` / `\d comments`)
- [ ] Existing Data Is Preserved (No Timestamps Shifted By Migration)
- [ ] Mobile App Still Displays Correct "Time Ago" Values Without The `Z`-Append Workaround From #217
- [ ] All Tests Pass

## Deadline

14.05.2026

## Notes

- Mobile Workaround Shipped In #217 — Can Be Cleaned Up In A Follow-Up Mobile Issue Once This Backend Fix Is Deployed.
- Related Files Already Identified: `backend/src/main/java/com/bounswe2026group1/backend/model/Report.java:45`, `model/Comment.java:32`, `dto/ReportResponse.java:27`, `dto/RampReportResponse.java:21`.

efemantarci:
# 📝 Use Instant Instead Of LocalDateTime For API Timestamps
Fixes #221

Switches `Report.publishDate` and `Comment.createdAt` (and the matching DTOs) from `LocalDateTime` to `Instant` so Jackson emits Z-suffixed ISO-8601 timestamps. An idempotent `schema.sql` migrates existing `TIMESTAMP` columns to `TIMESTAMPTZ`, reinterpreting stored values as UTC to match the production JVM timezone.

# 📊 Type of Change
- [x] Bug fix
- [x] Refactoring

- [x] All tests passed

# 📸 Screenshots / Recordings
N/A — wire-format change verified via Hibernate schema log: `publish_date timestamp(6) with time zone`, `created_at timestamp(6) with time zone`.

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/src/main/java/com/bounswe2026group1/backend/dto/RampReportResponse.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/dto/RampReportResponse.java
@@ -6,7 +6,7 @@ import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
-import java.time.LocalDateTime;
+import java.time.Instant;
 
 @Data
 @NoArgsConstructor
@@ -18,7 +18,7 @@ public class RampReportResponse {
     private double longitude;
     private String description;
     private ReportStatus status;
-    private LocalDateTime publishDate;
+    private Instant publishDate;
     private double entryLatitude;
     private double entryLongitude;
     private double exitLatitude;
--- a/backend/src/main/java/com/bounswe2026group1/backend/dto/ReportResponse.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/dto/ReportResponse.java
@@ -9,7 +9,7 @@ import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
-import java.time.LocalDateTime;
+import java.time.Instant;
 import java.util.List;
 
 @Data
@@ -25,7 +25,7 @@ public class ReportResponse {
     private ReportStatus status;
     private int agrees;
     private int disagrees;
-    private LocalDateTime publishDate;
+    private Instant publishDate;
     private List<String> mediaUrls;
     private VoteType userVote;
 
--- a/backend/src/main/java/com/bounswe2026group1/backend/model/Comment.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/model/Comment.java
@@ -6,7 +6,7 @@ import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
-import java.time.LocalDateTime;
+import java.time.Instant;
 
 @Entity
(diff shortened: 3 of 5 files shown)


SimsekMucahit:
I reviewed the changes and there is no remainig LocalDateTime usage in backend. So, the API timezone ambiuity is resolved. Good job

SimsekMucahit (approved):

PR #395, opened 2026-05-01
[Backend] Search Users

### LINKED ISSUE #310: [Backend] Search Users
# [Backend] Search Users

Parent issue: #TBD

## Description

Implement backend support for the user search feature in MapCess. The Spring Boot backend must expose a search endpoint that allows querying users by display name or username, returning paginated results with only publicly visible profile fields.

## Feature / Task Details

- Implement REST endpoint:
  - `GET /api/users/search?q={query}&page={page}&size={size}` — search users by display name or username (case-insensitive, partial match)
- Response per user item: `id`, `username`, `displayName`, `avatarUrl`, `contributionCount` (report submissions)
- Results must be paginated (default page size: 20)
- Only publicly visible fields should be returned — never expose email, preferences, or other private data
- The endpoint should be accessible to authenticated users only
- Add a database index on `username` and `displayName` columns for query performance
- Write unit and integration tests for the search endpoint

## Deliverables

- `GET /api/users/search` endpoint with partial-match, case-insensitive search and pagination
- DB index on searchable fields
- Unit and integration tests

## Acceptance Criteria

- [ ] `GET /api/users/search?q=...` returns a paginated list of matching users
- [ ] Search is case-insensitive and matches partial display name or username
- [ ] Response includes only public fields (id, username, displayName, avatarUrl, contributionCount)
- [ ] Private fields (email, preferences) are never returned
- [ ] Endpoint requires authentication
- [ ] Pagination works correctly (page, size, totalPages in response)
- [ ] DB indexes added for search performance
- [ ] Unit and integration tests written and passing
- [ ] API documented (Swagger/OpenAPI updated)

## Deadline

01.05.2026

## Notes

Agree on the API contract (query param names, response shape, pagination format) with the web and mobile teams before implementation. Consider using Spring Data JPA `Containing` / `IgnoreCase` queries or a `LIKE` query for partial matching.

efemantarci:
# 📝 Search Users
Fixes #310

Adds `GET /api/users/search?q=&page=&size=` — case-insensitive partial match on the user's `name` field, returning paginated `UserSearchDto` items (`id`, `name`, `avatarUrl`, `contributionCount`).

**Stacked on #389** (`backend/user-profile`) — that PR adds `avatarUrl` on `RegisteredUser` and `ReportRepository.countByCreatedById`, both required here. **Merge #389 first**, then retarget this PR's base to `main` and merge.

**Contract deviation from #310 (agreed up front):** the issue spec asks for `username` and `displayName` fields, but neither exists in the data model — #389 deliberately defers a separate username/handle field. We search and return the existing `name` field only. The DTO is additive: a separate `displayName`/`username` can be added later without breaking clients.

- New `findByNameContainingIgnoreCase(String, Pageable)` derived query, backed by a new B-tree index `idx_registered_users_name`.
- Service normalizes a null/whitespace `q` to `""` (matches all) and pairs each user in the page with their `reportRepository.countByCreatedById` — bounded by the default page size (20), so no N+1 at scale.
- Controller returns Spring's `Page<UserSearchDto>` directly. Defaults to `page=0`, `size=20` via `@PageableDefault`. Auth handled by `SecurityConfig.anyRequest().authenticated()` — no security config change.
- 5 repo tests (`@DataJpaTest`), 4 service tests, 4 controller tests (`@WebMvcTest`). Full backend suite: 111 passed, 0 failed, 1 skipped (live ORS).

# 📊 Type of Change
- [ ]  Bug fix
- [x]  New feature
- [ ]  Refactoring
- [ ]  Documentation

- [x] All tests passed

# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/src/main/java/com/bounswe2026group1/backend/dto/UserSearchDto.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/dto/UserSearchDto.java
@@ -1,0 +1,25 @@
+package com.bounswe2026group1.backend.dto;
+
+import com.bounswe2026group1.backend.model.RegisteredUser;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class UserSearchDto {
+    private Long id;
+    private String name;
+    private String avatarUrl;
+    private long contributionCount;
+
+    public static UserSearchDto fromEntity(RegisteredUser user, long contributionCount) {
+        return new UserSearchDto(
+                user.getId(),
+                user.getName(),
+                user.getAvatarUrl(),
+                contributionCount
+        );
+    }
+}
--- a/backend/src/main/java/com/bounswe2026group1/backend/dto/UserSearchDto.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/dto/UserSearchDto.java
@@ -12,7 +12,7 @@ public class UserSearchDto {
     private Long id;
     private String name;
     private String avatarUrl;
-    private long contributionCount;
+    private long reportCount;
 
     public static UserSearchDto fromEntity(RegisteredUser user, long contributionCount) {
         return new UserSearchDto(
--- a/backend/src/main/java/com/bounswe2026group1/backend/repository/RegisteredUserRepository.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/repository/RegisteredUserRepository.java
@@ -21,15 +21,15 @@ public interface RegisteredUserRepository extends JpaRepository<RegisteredUser, 
      */
     @Query(value = """
             SELECT u FROM RegisteredUser u
-            WHERE LOWER(u.name) LIKE LOWER(CONCAT('%', :q, '%'))
+            WHERE LOWER(u.name) LIKE LOWER(CONCAT('%', :q, '%')) ESCAPE '!'
             ORDER BY
-                CASE WHEN LOWER(u.name) LIKE LOWER(CONCAT(:q, '%')) THEN 0 ELSE 1 END,
+                CASE WHEN LOWER(u.name) LIKE LOWER(CONCAT(:q, '%')) ESCAPE '!' THEN 0 ELSE 1 END,
                 LOWER(u.name) ASC,
                 u.id ASC
             """,
             countQuery = """
             SELECT COUNT(u) FROM RegisteredUser u
-            WHERE LOWER(u.name) LIKE LOWER(CONCAT('%', :q, '%'))
+            WHERE LOWER(u.name) LIKE LOWER(CONCAT('%', :q, '%')) ESCAPE '!'
             """)
     Page<RegisteredUser> searchByName(@Param("q") String q, Pageable pageable);
 }
--- a/backend/src/main/java/com/bounswe2026group1/backend/repository/RegisteredUserRepository.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/repository/RegisteredUserRepository.java
@@ -1,6 +1,8 @@
 package com.bounswe2026group1.backend.repository;
 
 import com.bounswe2026group1.backend.model.RegisteredUser;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 
@@ -8,4 +10,6 @@ import org.springframework.stereotype.Repository;
 public interface RegisteredUserRepository extends JpaRepository<RegisteredUser, Long> {
     boolean existsByEmail(String email);
     java.util.Optional<RegisteredUser> findByEmail(String email);
+
+    Page<RegisteredUser> findByNameContainingIgnoreCase(String name, Pageable pageable);
 }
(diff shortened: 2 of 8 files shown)


backend/src/main/java/com/bounswe2026group1/backend/service/RegisteredUserService.java
```diff
@@ -153,30 +129,34 @@ public UserProfileDTO updateProfile(Long id, UpdateProfileRequest request) {
...
+    }
+
+    // ───── Search (issue #310) ──────────────────────────────────────────────
+
+    public Page<UserSearchDto> searchUsers(String q, Pageable pageable) {
+        String query = q == null ? "" : q.trim();
+        return registeredUserRepository
+                .searchByName(query, pageable)
+                .map(u -> UserSearchDto.fromEntity(
+                        u, reportRepository.countByCreatedById(u.getId())));
```
alaaddinerenn: `searchUsers` issues N+1 COUNT queries — one per user in the page.`reportRepository.countByCreatedById(u.getId())` is called inside `.map()` for every user returned by `searchByName`. For a page of 20, that is 20 separate round-trips. Please add a batch count query, e.g.:
`@Query("SELECT r.createdBy.id, COUNT(r) FROM Report r WHERE r.createdBy.id IN :ids GROUP BY r.createdBy.id")`
and resolve the map in the service before building the DTOs.

backend/src/main/java/com/bounswe2026group1/backend/repository/RegisteredUserRepository.java
```diff
@@ -1,11 +1,35 @@
...
+
+    /**
+     * Case-insensitive name search ordered by relevance:
+     * names that START WITH the query come before names that merely CONTAIN it,
+     * with each group sorted alphabetically (case-insensitive) and id as a
+     * deterministic tiebreaker. An empty query matches every user.
+     */
+    @Query(value = """
+            SELECT u FROM RegisteredUser u
+            WHERE LOWER(u.name) LIKE LOWER(CONCAT('%', :q, '%'))
```
alaaddinerenn: LIKE wildcard characters in `q` are not escaped. The search query uses `LIKE CONCAT('%', :q, '%')`. If a user submits `%` or `_`, those characters are interpreted as SQL LIKE wildcards (`%` matches everything, `_` matches any single character), producing unintended results.

Please escape `\`, `%`, and `_` in the trimmed query before passing it to the repository, and add `ESCAPE '\\'` to both LIKE clauses in the JPQL.

backend/src/main/java/com/bounswe2026group1/backend/service/RegisteredUserService.java
```diff
@@ -153,30 +129,34 @@ public UserProfileDTO updateProfile(Long id, UpdateProfileRequest request) {
         }
 
         RegisteredUser saved = registeredUserRepository.save(user);
-        return toProfileDTO(saved, true);
+        return toProfileDTO(saved);
+    }
+
+    // ───── Search (issue #310) ──────────────────────────────────────────────
+
+    public Page<UserSearchDto> searchUsers(String q, Pageable pageable) {
```
alaaddinerenn, line 167: Add `@Transactional(readOnly = true)` to `searchUsers`. The method issues multiple DB queries (searchByName + count per user) with no transaction boundary. Annotating it `@Transactional(readOnly = true)` avoids Hibernate dirty-check overhead and signals to the connection pool that no writes will occur.

backend/src/main/java/com/bounswe2026group1/backend/dto/UserSearchDto.java
```diff
@@ -0,0 +1,25 @@
...
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class UserSearchDto {
+    private Long id;
+    private String name;
+    private String avatarUrl;
+    private long contributionCount;
```
alaaddinerenn: `contributionCount` should either include routes or be renamed. `UserProfileDTO` exposes separate `reportsSubmitted` and `routesPlanned` counters. `UserSearchDto.contributionCount` only counts reports, silently ignoring routes. Either rename it to `reportCount` to be precise, or sum both report and route counts to match what `UserProfileDTO` represents as "contributions".

alaaddinerenn (changes requested):
It would be great if you could take care of the requested modifications at your convenience. @efemantarci

backend/src/main/java/com/bounswe2026group1/backend/dto/UserSearchDto.java
efemantarci: contributionCount is renamed to reportCount now

backend/src/main/java/com/bounswe2026group1/backend/service/RegisteredUserService.java
efemantarci: @Transactional(readOnly = true) added to searchUsers

backend/src/main/java/com/bounswe2026group1/backend/repository/RegisteredUserRepository.java
efemantarci: LIKE wildcards are escaped with ! before reaching the repo now

backend/src/main/java/com/bounswe2026group1/backend/service/RegisteredUserService.java
efemantarci: searchUsers now fetches all report counts in a single batch query (IN :ids GROUP BY) instead of one per user

alaaddinerenn (approved):
All changes look good — thanks for the quick turnaround on the feedback. Approved!

PR #405, opened 2026-05-03
[Backend][Bug] Fix Backend Startup Failures In Docker Local Compose

### LINKED ISSUE #404: [Backend] Schema.sql Dollar-Quote Block Crashes Backend In Docker Local Compose
# 📝 Schema.sql Dollar-Quote Block Crashes Backend In Docker Local Compose

## Description
Running `docker compose -f docker-compose.local.yml up --build -d` fails to start the `mapcess_backend_dev` container. The error is thrown from `dataSourceScriptDatabaseInitializer` while loading `backend/src/main/resources/schema.sql`:

```
Caused by: org.postgresql.util.PSQLException: Unterminated dollar quote started at position 3 in SQL DO $$ BEGIN IF EXISTS ( SELECT 1 FROM information_schema.columns WHERE table_name = 'reports' AND column_name = 'publish_date' AND data_type = 'timestamp without time zone' ) THEN ALTER TABLE reports ALTER COLUMN publish_date TYPE TIMESTAMPTZ USING publish_date AT TIME ZONE 'UTC'. Expected terminating $$
```

Spring's `ScriptUtils` (used by `DataSourceScriptDatabaseInitializer`) splits SQL scripts on `;` and only recognizes `'...'` and `"..."` quoting — it has no understanding of PostgreSQL dollar-quoting (`$$ ... $$`). The migration in `schema.sql` (introduced in #394, commit `813d9c2`) wraps multiple `ALTER TABLE` statements inside a `DO $$ BEGIN ... END $$;` block. Spring slices the file at the first `;` it encounters inside the block (after `'UTC';`), and the resulting fragment fails to parse because the `$$` is unclosed. The migration logic itself is valid PostgreSQL — only the Spring-side script loader can't handle it.

The bug ships because the only execution path that triggers it is `docker compose -f docker-compose.local.yml up`, and that path isn't covered by CI or by the most common dev workflows: `backend-pr-check.yml` runs `./mvnw verify` against H2 with `spring.sql.init.mode=never` (set in `application-test.properties`); `application.properties` defaults to `never` so `./mvnw spring-boot:run` doesn't trigger it; only the local compose explicitly sets `SPRING_SQL_INIT_MODE=always`.

## Feature / Task Details
Three workable fixes, ranked by invasiveness:

- **Option 1 — Drop the `DO $$ ... $$;` block from `schema.sql`.** Acceptable because (a) prod was migrated when #394 landed, (b) fresh local DBs get `TIMESTAMPTZ` from Hibernate's `ddl-auto=update` since `Report.publishDate` and `Comment.createdAt` are `Instant`. Lowest-risk; loses the migration safety net for devs with stale local DBs (mitigation: `docker compose down -v` resets).
- **Option 2 — Configure `spring.sql.init.separator=^---^`** (or any unique sentinel) and update both `schema.sql` and `data.sql` to use it as the statement terminator. Preserves the migration; touches ~30 statement terminators in `data.sql`.
- **Option 3 — Move the migration to a Java `@Bean ApplicationRunner`** that opens a JDBC `Statement` and `execute()`s the DO block as one string. Bypasses Spring's splitter entirely. Heaviest but most future-proof; opens the door to Flyway/Liquibase later.

Recommendation: ship Option 1 to unblock local testing now; file a follow-up to introduce Flyway if column-type migrations become a recurring pattern.

## Deliverables
- A `schema.sql` that does not crash at startup with `SPRING_SQL_INIT_MODE=always`.
- If keeping the migration: working timestamp migration for legacy DBs, or a documented manual `psql` script if the migration is removed.
- Backend health verified end-to-end via `docker compose -f docker-compose.local.yml up --build -d`.

## Acceptance Criteria

- [ ] `docker compose -f docker-compose.local.yml up --build -d` starts the backend cleanly with no `Unterminated dollar quote` error.
- [ ] Backend `GET /api/reports` returns `200` after startup.
- [ ] `data.sql` seed still loads correctly (Adminer shows seeded users + reports).
- [ ] All tests pass.

## Deadline
2026-05-10

## Notes
- Refs: #221 (introduced the timestamp-type change), #394 (PR that merged the broken `schema.sql`).
- Discovered while testing #401 (web fix-flow UI) locally — first time anyone re-spun the local compose after #394 merged.

### LINKED ISSUE #406: [Backend] Data.sql Comment Inserts Reference Non-Existent r.id Column
# 📝 Data.sql Comment Inserts Reference Non-Existent r.id Column

## Description
Backend startup fails when `data.sql` runs (`SPRING_SQL_INIT_MODE=always`, set by `docker-compose.local.yml`). Statement #8 — the first comment INSERT — throws:

```
Caused by: org.postgresql.util.PSQLException: ERROR: column r.id does not exist
  Position: 116
```

The seed data references `r.id` for the `reports` alias in three comment-insert subqueries (six total occurrences across the WHERE NOT EXISTS guards and the SELECT projections). The `reports` table does not have an `id` column — its primary key column is `report_id`, because [Report.java](backend/src/main/java/com/bounswe2026group1/backend/model/Report.java) names the field `reportId` and Hibernate maps it via snake_case.

Same root cause as #404: PR #384 (`fix/data-sql-fk-safety`) was merged today without anyone running `docker compose -f docker-compose.local.yml up`. CI doesn't catch it because tests use H2 with `spring.sql.init.mode=never`, and `application.properties` defaults to `never` too. Only the local compose triggers the broken path.

## Feature / Task Details
Replace all six occurrences of `r.id` with `r.report_id` in `backend/src/main/resources/data.sql` (lines 69, 74, 78, 83, 87, 92). All other column references in `data.sql` were audited and are correct.

## Deliverables
- A `data.sql` whose comment INSERTs successfully reference the parent reports table.
- Backend boots cleanly through `data.sql` with `SPRING_SQL_INIT_MODE=always`.
- Seeded comments visible in Adminer after first startup.

## Acceptance Criteria

- [ ] `docker compose -f docker-compose.local.yml up --build -d` completes with backend healthy.
- [ ] `SELECT * FROM comments;` in Adminer shows the three seeded rows.
- [ ] CI continues to pass (`./mvnw verify`).
- [ ] All tests pass.

## Deadline
2026-05-10

## Notes
- Refs: #384 (PR that introduced the comment-insert bug), #404 (sibling startup-failure bug).
- Discovered while testing #401 fix-flow UI locally, immediately after #404's schema.sql fix unblocked startup past the schema-init phase.

efemantarci:
# 📝 Fix Backend Startup Failures In Docker Local Compose
Fixes #404
Fixes #406

Two pre-existing bugs were preventing `docker compose -f docker-compose.local.yml up --build -d` from booting the backend. CI never caught them because it runs `./mvnw verify` against H2 with `spring.sql.init.mode=never`; `application.properties` defaults to `never` too, so only the local compose's `SPRING_SQL_INIT_MODE=always` triggers the broken paths. Both bugs were merged in PRs landed earlier today without anyone re-spinning the local stack.

This PR ships three small commits:

1. **`fix(db): empty schema.sql to unblock docker-local startup`** — drops the PostgreSQL `DO $$ … END $$;` block that Spring's `ScriptUtils` couldn't parse (it has no awareness of dollar-quoting and was slicing the block mid-quote). Fresh DBs still get `TIMESTAMPTZ` from Hibernate's `ddl-auto=update` since the relevant entity fields are `Instant`. (#404)
2. **`fix(db): keep schema.sql non-empty after stripping comments`** — Spring strips comments before validating the script, then rejects an empty result with `IllegalArgumentException: 'script' must not be null or empty`. Adds a harmless `SELECT 1;` no-op. (#404)
3. **`fix(db): replace r.id with r.report_id in data.sql comment inserts`** — the comment-insert subqueries aliased `reports` as `r` and referenced `r.id`, but the PK column is `report_id` (Report.reportId → snake_case). Replaces six occurrences. (#406)

# 📊 Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation

- [x] All tests passed

# 📸 Screenshots / Recordings
N/A — server-side init scripts.

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/src/main/resources/data.sql
+++ b/backend/src/main/resources/data.sql
@@ -66,28 +66,28 @@ SELECT setval('report_categories_id_seq', (SELECT MAX(id) FROM report_categories
 -- referenced parent report cannot be found.
 -- Comments: guard with NOT EXISTS on content + report_id
 INSERT INTO comments (content, author_id, report_id, created_at)
-SELECT 'Yes, people also fall from this ramp', 2, r.id, NOW()
+SELECT 'Yes, people also fall from this ramp', 2, r.report_id, NOW()
 FROM reports r
 WHERE r.description = 'This ramp is too steep for wheelchairs'
   AND NOT EXISTS (
     SELECT 1 FROM comments c
-    WHERE c.content = 'Yes, people also fall from this ramp' AND c.report_id = r.id
+    WHERE c.content = 'Yes, people also fall from this ramp' AND c.report_id = r.report_id
   );
 
 INSERT INTO comments (content, author_id, report_id, created_at)
-SELECT 'Please fix this ASAP, it affects many students.', 3, r.id, NOW()
+SELECT 'Please fix this ASAP, it affects many students.', 3, r.report_id, NOW()
 FROM reports r
 WHERE r.description = 'This ramp is too steep for wheelchairs'
   AND NOT EXISTS (
     SELECT 1 FROM comments c
-    WHERE c.content = 'Please fix this ASAP, it affects many students.' AND c.report_id = r.id
+    WHERE c.content = 'Please fix this ASAP, it affects many students.' AND c.report_id = r.report_id
   );
 
 INSERT INTO comments (content, author_id, report_id, created_at)
-SELECT 'The elevator was repaired according to staff.', 3, r.id, NOW()
+SELECT 'The elevator was repaired according to staff.', 3, r.report_id, NOW()
 FROM reports r
 WHERE r.description = 'Elevator of Boğaziçi Metro is broken'
   AND NOT EXISTS (
     SELECT 1 FROM comments c
-    WHERE c.content = 'The elevator was repaired according to staff.' AND c.report_id = r.id
+    WHERE c.content = 'The elevator was repaired according to staff.' AND c.report_id = r.report_id
   );
--- a/backend/src/main/resources/schema.sql
+++ b/backend/src/main/resources/schema.sql
@@ -1,32 +1,3 @@
--- Migrate legacy timestamp columns to TIMESTAMPTZ so the API emits Z-suffixed
--- timestamps via Jackson's Instant serializer. See issue #221.
---
--- Runs after Hibernate (spring.jpa.defer-datasource-initialization=true) and
--- before data.sql. Idempotent: a no-op once the column is already TIMESTAMPTZ.
--- Existing values are reinterpreted as UTC, which matches the production JVM
--- timezone (containers run in UTC) so wall-clock instants are preserved.
-
-DO $$
-BEGIN
-    IF EXISTS (
-        SELECT 1 FROM information_schema.columns
-        WHERE table_name = 'reports'
-          AND column_name = 'publish_date'
-          AND data_type = 'timestamp without time zone'
-    ) THEN
-        ALTER TABLE reports
-            ALTER COLUMN publish_date TYPE TIMESTAMPTZ
-            USING publish_date AT TIME ZONE 'UTC';
-    END IF;
-
-    IF EXISTS (
-        SELECT 1 FROM information_schema.columns
-        WHERE table_name = 'comments'
-          AND column_name = 'created_at'
-          AND data_type = 'timestamp without time zone'
-    ) THEN
-        ALTER TABLE comments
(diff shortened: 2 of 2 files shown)


alaaddinerenn (approved):
Tested locally with Docker — the project builds and runs successfully. Approved!

PR #423, opened 2026-05-08
[Backend] Accessibility Profiles and Constraint-Based Routing

### LINKED ISSUE #365: [Backend] Accessibility Profiles and Constraint-Based Routing

efemantarci:
# 📝 Accessibility Profiles and Constraint-Based Routing
Fixes #365

Adds backend infrastructure for stored user accessibility preferences and
threads those preferences through the routing pipeline. Users can pick a named
preset (`WHEELCHAIR_USER`, `BLIND_OR_LOW_VISION`, `MOBILITY_LIMITED`) or a
custom mix of 10 path-level constraints (`AVOID_STAIRS`, `REQUIRE_RAMPS`,
`AVOID_NARROW_SIDEWALKS`, `AVOID_DAMAGED_TACTILE_PAVING`, etc.); the avoid
polygons fed to OpenRouteService now include only obstacles whose hazard
tuples match the caller's constraint set, and the matching alternative on the
3-route response is flagged `preferred:true`. Anonymous and `NONE`-preset
callers see the existing baseline (every VERIFIED OUTDOOR obstacle becomes a
polygon).

The PR also fixes three pre-existing routing bugs surfaced by hands-on testing
of the feature:

1. **Native-query enum binding** — `findByTypeInBoundingBoxWithStatuses` and
   `findReportsInBoundingBoxWithStatuses` were 500-ing prod (`operator does
   not exist: character varying = smallint`) because Hibernate binds
   `Collection<Enum>` as ordinals in native queries. Switched the params to
   `Collection<String>` and the single caller (`ObstacleService`) now passes
   `ReportStatus.VERIFIED.name()`.
2. **INDOOR reports leaking into routing** — `buildAvoidPolygons` was
   generating polygons for indoor reports (e.g. broken elevators inside
   buildings) which ORS foot-walking can't navigate to anyway. Filtered all
   three routing repository queries to `environment = OUTDOOR`. Indoor
   reports stay in the report system as informational data.
3. **`hasObstacles` flag ignored constraints** — a `⭐ preferred` Fastest
   Route could warn about an obstacle the user had explicitly declared they
   don't care about. `findObstaclesOnPath` now applies the same
   constraint-filter predicate as `buildAvoidPolygons`.

# 📊 Type of Change
- [x] Bug fix (3 pre-existing bugs surfaced by testing the new feature)
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

- [x] All tests passed

## What's tested
- **44 new tests** across 7 files (313/313 in `mvnw clean verify`, 1 skipped live-ORS):
  - 17 unit tests on the enums (constraint→hazard mapping, preset round-trip)
  - 15 unit tests on `RoutingPreferencesService` (every PUT permutation)
  - 5 new unit tests on `ObstacleService` (constraint-aware avoid polygons)
  - 14 PostGIS-backed integration tests (`ObstacleServiceConstraintIntegrationTest`)
    including 2 native-query enum-binding regression guards, 1 INDOOR-filter
    guard, and 3 constraint-aware `findObstaclesOnPath` guards
  - 10 full HTTP integration tests on the prefs endpoints (`@SpringBootTest` +
    real DB) covering anonymous 401, fresh-user catalog, every PUT semantic,
    PUT-then-GET round-trip, toggle workflow with auto-detected preset
  - 6 end-to-end tests confirming user constraints flow into ObstacleService
    via `ArgumentCaptor` and the `preferred:true` flag is set on the right alt
- **Live smoke test** against the local Docker stack: PUT prefs through curl,
  verified DB persistence in Adminer, and confirmed `/api/routes` deflects the
  fastest path (270m → 535m) when the user's constraints match the seeded
  blocked-sidewalk obstacle, and stays unchanged when they don't.

# 📸 Screenshots / Recordings
N/A — backend-only changes.

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [ ] 5-15 min
- [x] > 15 min

### DIFF
--- a/backend/src/main/java/com/bounswe2026group1/backend/model/RoutingConstraint.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/model/RoutingConstraint.java
@@ -1,0 +1,132 @@
+package com.bounswe2026group1.backend.model;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static com.bounswe2026group1.backend.model.IssueType.*;
+import static com.bounswe2026group1.backend.model.ObjectType.*;
+
+/**
+ * High-level user-facing routing constraints. Each value carries the set of
+ * (ObjectType, IssueType) hazard tuples it covers. The avoid-polygon builder
+ * unions hazards across the user's selected constraints to decide which
+ * VERIFIED reports become avoidance polygons.
+ *
+ * Constraints intentionally cover only path-segment hazards (sidewalks, ramps,
+ * stairs, low overhead clearance). Door and elevator issues are destination-level
+ * and stay informational — they don't generate avoid polygons.
+ */
+public enum RoutingConstraint {
+
+    AVOID_STAIRS(
+            "Avoid stairs entirely",
+            "Routes will avoid sidewalk segments containing reported stairways.",
+            Set.of(
+                    new IssueHazard(STAIR, MISSING_HANDRAIL),
+                    new IssueHazard(STAIR, NO_LANDING),
+                    new IssueHazard(STAIR, TOO_NARROW),
+                    new IssueHazard(STAIR, SLIPPERY_SURFACE),
+                    new IssueHazard(STAIR, RISER_TOO_HIGH),
+                    new IssueHazard(STAIR, TREAD_TOO_SHALLOW),
+                    new IssueHazard(STAIR, NO_ANTI_SLIP),
+                    new IssueHazard(STAIR, OPEN_RISERS))),
+
+    REQUIRE_RAMPS(
+            "Prefer routes with ramps over steps",
+            "Treats reports of missing ramps as obstacles to route around.",
+            Set.of(new IssueHazard(RAMP, MISSING))),
+
+    AVOID_UNSAFE_RAMPS(
+            "Avoid unsafe ramps",
+            "Routes will avoid ramps reported as too steep, too narrow, missing handrails, missing landings, or slippery.",
+            Set.of(
+                    new IssueHazard(RAMP, TOO_STEEP),
+                    new IssueHazard(RAMP, TOO_NARROW),
+                    new IssueHazard(RAMP, NO_LANDING),
+                    new IssueHazard(RAMP, SLIPPERY_SURFACE),
+                    new IssueHazard(RAMP, MISSING_HANDRAIL))),
+
+    AVOID_NARROW_SIDEWALKS(
+            "Avoid narrow sidewalks",
+            "Routes will avoid sidewalks reported as too narrow for accessible passage.",
+            Set.of(new IssueHazard(SIDEWALK, TOO_NARROW))),
+
+    AVOID_LOW_CLEARANCE(
+            "Avoid low overhead clearance",
+            "Routes will avoid sidewalk segments with low-hanging branches, signs, or other overhead obstructions.",
+            Set.of(new IssueHazard(SIDEWALK, INSUFFICIENT_CLEARANCE))),
+
+    AVOID_SLIPPERY_SURFACES(
+            "Avoid slippery surfaces",
+            "Routes will avoid ramps, sidewalks, and stairs reported as slippery or lacking anti-slip surfaces.",
+            Set.of(
+                    new IssueHazard(RAMP, SLIPPERY_SURFACE),
+                    new IssueHazard(SIDEWALK, SLIPPERY_SURFACE),
+                    new IssueHazard(STAIR, SLIPPERY_SURFACE),
+                    new IssueHazard(STAIR, NO_ANTI_SLIP))),
+
+    REQUIRE_HANDRAILS(
+            "Require handrails on stairs and ramps",
+            "Routes will avoid stairs and ramps reported as missing handrails.",
+            Set.of(
+                    new IssueHazard(RAMP, MISSING_HANDRAIL),
+                    new IssueHazard(STAIR, MISSING_HANDRAIL))),
+
+    AVOID_BLOCKED_OR_MISSING_SIDEWALKS(
+            "Avoid blocked or missing sidewalks",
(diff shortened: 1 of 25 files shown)


backend/src/main/java/com/bounswe2026group1/backend/service/RoutingPreferencesService.java
```diff
@@ -0,0 +1,127 @@
...
+
+        Set<RoutingConstraint> resolvedConstraints;
+        if (request.getConstraints() != null) {
+            // Constraints win when both fields are sent — see UpdateRoutingPreferencesRequest contract.
+            resolvedConstraints = sanitize(request.getConstraints());
+        } else if (request.getPreferredPreset() != null) {
+            resolvedConstraints = EnumSet.copyOf(
+                    request.getPreferredPreset() == RoutingPreset.CUSTOM
+                            ? user.getRoutingConstraints()
+                            : request.getPreferredPreset().getConstraints());
```
alaaddinerenn: RoutingPreferencesService.update (lines 62–65) will throw IllegalArgumentException when a request with preferredPreset = CUSTOM arrives for a user whose routingConstraints set is empty. user.getRoutingConstraints() is initialized as a HashSet (or wrapped by Hibernate as PersistentSet), and EnumSet.copyOf(Collection) rejects non-EnumSet empty collections with IllegalArgumentException. Please guard the CUSTOM branch — for example: EnumSet<RoutingConstraint> existing = user.getRoutingConstraints().isEmpty() ? EnumSet.noneOf(RoutingConstraint.class) : EnumSet.copyOf(user.getRoutingConstraints()); — and add a regression test covering a fresh user (no constraints) sending {"preferredPreset":"CUSTOM"}. The current update_withCustomPresetOnly_keepsExistingConstraintsAndDetectsCustom test only exercises the non-empty case, which is why this bug isn't caught.

backend/src/main/java/com/bounswe2026group1/backend/model/RoutingConstraint.java
```diff
@@ -0,0 +1,132 @@
...
+                    new IssueHazard(STAIR, RISER_TOO_HIGH),
+                    new IssueHazard(STAIR, TREAD_TOO_SHALLOW),
+                    new IssueHazard(STAIR, NO_LANDING))),
+
+    AVOID_DAMAGED_TACTILE_PAVING(
+            "Avoid damaged tactile paving",
+            "Routes will avoid sidewalks with reported tactile-paving issues that may misdirect visually impaired users.",
+            // Currently maps to NO_TACTILE_PAVING. The IssueType will be renamed to
+            // TACTILE_PAVING_DAMAGED in a follow-up; only this hazard set updates.
+            Set.of(new IssueHazard(SIDEWALK, NO_TACTILE_PAVING)));
```
alaaddinerenn: The AVOID_DAMAGED_TACTILE_PAVING constraint is labeled "Avoid damaged tactile paving" but its hazard tuple is (SIDEWALK, NO_TACTILE_PAVING), which means missing tactile paving, not damaged. A blind or low-vision user selecting this preference will reasonably expect routes to avoid sidewalks with broken/worn tactile paving, but they'll actually get routes avoiding sidewalks that have no tactile paving at all — a meaningfully different outcome. Please resolve the mismatch in this PR rather than deferring: either (a) rename the enum value to AVOID_MISSING_TACTILE_PAVING and update the label/description to "Avoid sidewalks lacking tactile paving" so semantics match the current hazard mapping, or (b) introduce a new TACTILE_PAVING_DAMAGED IssueType and bind it here. Shipping the current state would mislead the exact user group this constraint targets.

alaaddinerenn (changes requested):
Overall great work! Just a couple of things I'd like you to address before we merge — could you take care of the changes I mentioned? Happy to approve once those are in.

backend/src/main/java/com/bounswe2026group1/backend/service/RoutingPreferencesService.java
efemantarci: You are right. I solved it and added an additional test.

backend/src/main/java/com/bounswe2026group1/backend/model/RoutingConstraint.java
efemantarci: I solved with a commit here. But I opened issue #430 I think damaged paving might be a better and realistic issue.

alaaddinerenn (approved):
Left some comments earlier and you've handled them all — good stuff. Code is clean, logic is solid. Approved and ready to merge.

PR #407, opened 2026-05-03
[Web] Add Fix Flow UI To Report Detail Panel

efemantarci:
# 📝 Add Fix Flow UI To Report Detail Panel
Fixes #401

Wires the web report-detail panel into the new fix-request endpoints (#403): a "Has this been fixed?" CTA for authenticated users, a modal sheet for submitting a fix report (photo + optional description), and a pinned green-themed active-fix card with its own agree/disagree vote bar. The fix card sits at the very top of the panel so a returning voter sees the live question first; the original report sits below as evidence. Status pills now cover `Verified / Unverified / Fixed / Rejected` plus a sub-pill for "Fix pending" when an OPEN fix request is in flight. SSE handler refetches the list cache on `fix_requested / fix_updated / fixed` ops so consensus updates live across browsers.

> Base branch is `backend/fix-status-validation-flow` (PR #403). GitHub will retarget to `main` automatically when #403 merges. Diff stays clean in the meantime.

# 📊 Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# ✅ Acceptance Criteria / Checklist
- [x] Project builds successfully
- [x] I have performed a self-review
- [x] I have commented my code, especially in hard-to-understand areas
- [x] I have added/updated tests
- [x] All tests passed

# 📸 Screenshots / Recordings
Tested locally end-to-end: submit fix → other browser sees `Fix pending` pill via SSE → 5th agree triggers `FIXED`, card disappears, status flips to `Fixed` live in both browsers.

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [ ] 5-15 min
- [x] > 15 min

PR #403, opened 2026-05-03
[Backend] Implement Fixed Status With Independent Validation Flow For Resolved Reports

efemantarci:
# 📝 Implement Fixed Status With Independent Validation Flow For Resolved Reports
Fixes #400

Adds the `FIXED` report status, a parallel fix-request voting flow (one OPEN per report at a time, requires media, votes independently of the main agree/disagree on the parent report), live SSE op tags `fix_requested` / `fixed`, and an hourly scheduled cleanup that expires stale OPEN fix requests after 7 days and deletes FIXED reports 7 days post-transition.

# 📊 Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

# ✅ Acceptance Criteria / Checklist
- [x] Project builds successfully
- [x] I have performed a self-review
- [x] I have commented my code, especially in hard-to-understand areas
- [x] I have added/updated tests
- [x] All tests passed

# 📸 Screenshots / Recordings
N/A — backend-only changes.

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [ ] 5-15 min
- [x] > 15 min

PR #426, opened 2026-05-08
[Web] Add Fix Flow UI To Report Detail Panel

delos0:
Retarget of #407 — now targeting main after #403 merged.


efemantarci (approved):
Nice catch on the revokeObjectURL cleanup. Merging

PR #443, opened 2026-05-08
[Backend][Bug] Verified Reports Don't Revert to Pending When Agrees Drop Below Threshold

### LINKED ISSUE #441: [Backend] Verified Reports Don't Revert to Pending When Agrees Drop Below Threshold
# 📝 Verified Reports Don't Revert to Pending When Agrees Drop Below Threshold

## Description

Once a report reaches `VERIFIED` (5 agrees, ≥60% ratio), it stays verified even if the agree count later drops below 5 due to vote retractions. The state machine in `evaluateStatus` only reverts a verified report when **disagrees ≥ 5**, never when agrees fall under the threshold. Symmetric problem on the `REJECTED` side.

How to recreate: vote agree on a pending report until it flips to `VERIFIED` with 5 agrees and 0 disagrees, then retract one agree. Status remains `VERIFIED` with only 4 agrees.

## Feature / Task Details

- File: `backend/src/main/java/com/bounswe2026group1/backend/service/ReportService.java`
- The fall-back-to-`PENDING` branches require `disagrees ≥ threshold` (or `agrees ≥ threshold` for the REJECTED case). They should also trip when the originally-asserted condition no longer holds.
- A status flip back to `PENDING` should also fan out a STATUS_CHANGE notification, same as the forward transition.

## Deliverables

- `evaluateStatus` reverts `VERIFIED → PENDING` when `agrees < threshold` or `agreeRatio < verificationRatio`
- `evaluateStatus` reverts `REJECTED → PENDING` when `disagrees < threshold` or `disagreeRatio < verificationRatio`
- Unit test covering the retraction case

## Acceptance Criteria

- [ ] Pulling an agree from a 5/0 verified report flips it back to PENDING
- [ ] Pulling a disagree from a 0/5 rejected report flips it back to PENDING
- [ ] Followers/voters receive a STATUS_CHANGE notification on the revert
- [ ] All tests pass

## Deadline
08.05.2026

## Notes

efemantarci:
# 📝 Verified Reports Don't Revert to Pending When Agrees Drop Below Threshold
Fixes #441

Once a report reached `VERIFIED`, `evaluateStatus` only reverted to `PENDING` when 5+ disagrees came in with ratio < 60%, leaving stale-criteria reports stuck verified after a vote retraction (e.g. 4 agrees, 0 disagrees, still VERIFIED). Same asymmetry on the `REJECTED` side. The fix collapses both revert branches into a single fall-through and extends the status-change fan-out so PENDING reverts produce a notification.

# 📊 Type of Change
- [x] Bug fix

- [x] All tests passed

# ⏱️ Estimated Review Time
- [x] < 5 min

### DIFF
--- a/backend/src/main/java/com/bounswe2026group1/backend/service/NotificationService.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/service/NotificationService.java
@@ -70,20 +70,18 @@ public class NotificationService {
         return saved;
     }
 
-    /** Trigger: report status transitioned to VERIFIED, REJECTED, or FIXED.
+    /** Trigger: report status transitioned to PENDING, VERIFIED, REJECTED, or FIXED.
      *  Notifies the report author, anyone who commented or voted on the report,
      *  and any explicit subscribers, minus the actor whose action triggered the
      *  change. Caller must only invoke this when the status actually changed. */
     public void notifyStatusChange(Report report, Long actorUserId) {
         if (report == null || report.getCreatedBy() == null) return;
         ReportStatus status = report.getStatus();
-        if (status != ReportStatus.VERIFIED
+        if (status != ReportStatus.PENDING
+                && status != ReportStatus.VERIFIED
                 && status != ReportStatus.REJECTED
                 && status != ReportStatus.FIXED) return;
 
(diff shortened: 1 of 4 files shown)


alaaddinerenn (approved):
Nice fixes mate, I have tested it manually and it works now!

PR #458, opened 2026-05-09
[Backend] Adopt Flyway with Enum Constraint Drop

### LINKED ISSUE #151: [Backend] Migrate from Hibernate Auto-DDL to a Database Migration Tool
# 📝 Migrate from Hibernate Auto-DDL to a Database Migration Tool

## Description
Currently, our Spring Boot application uses `spring.jpa.hibernate.ddl-auto=update` to manage database schemas. While this is useful for rapid prototyping, it poses a high risk for data loss and inconsistency in production environments. We need to disable Hibernate's automatic schema generation and implement a dedicated database migration tool to track, manage, and version our database schema changes safely.

## Feature / Task Details
- **Evaluate Alternatives:** Research and select the most suitable database schema management tool. Options to consider:
  - **Flyway:** SQL-based, straightforward, industry standard for Spring Boot.
  - **Liquibase:** Supports XML/YAML/JSON/SQL, more complex but highly flexible.
  - **Supabase / BaaS:** Evaluate if migrating the entire database infrastructure to a managed service like Supabase is necessary, or if we stick to our current Postgres setup.
- **Update Configuration:** Change the Hibernate setting to `spring.jpa.hibernate.ddl-auto=validate` (or `none`) in `application.properties` to prevent auto-updates.
- **Baseline Migration:** Generate the initial baseline migration script (e.g., `V1__init_schema.sql`) that reflects the current state of our database entities.
- **Integration:** Add the chosen tool's dependency to `pom.xml` and configure it to run automatically on application startup.

## Deliverables
- A brief comment/decision log on which tool was chosen and why.
- Updated `pom.xml` with the necessary dependencies.
- Updated `application.properties` with the new schema management configurations.
- The initial migration script file pushed to the repository.

## Acceptance Criteria
- [ ] Migration tool is selected and successfully integrated into the project.
- [ ] Hibernate auto-DDL is disabled in production configurations.
- [ ] The application successfully boots up and applies the baseline migration to a fresh, empty PostgreSQL database.
- [ ] All tests pass.
- [ ] Documentation updated (Add a README section explaining how other developers should write new database migration scripts in the future).

## Deadline
26.04.2026

## Notes
- Let's keep it simple for now. If we don't have complex needs, Flyway is usually the easiest to set up with Spring Boot since we can just write plain SQL. 
- Make sure to backup the existing development database before testing the first migration!

### LINKED ISSUE #456: [Backend] Drop legacy CHECK constraints blocking new enum values
# 📝 Drop Legacy CHECK Constraints Blocking New Enum Values

## Description
After expanding the report taxonomy (4 new ObjectTypes, 26 new IssueTypes), production POST /api/reports requests using any of the new enum values (WASHROOM, ROOM_SIGN, PEDESTRIAN_CROSSING, CURB_RAMP, and the new issue types) return 500 Internal Server Error. Inspection of pg_constraint shows that Hibernate's ddl-auto=update did not extend the original CHECK constraints on report_objects.object_type and report_object_issues.issue_type when the new enum values were added, so the DB still rejects them at insert time. The application layer validates the enums via @Enumerated(EnumType.STRING), so DB-level CHECK constraints are redundant and create drift every time the enum set grows.

## Feature / Task Details

- Add idempotent ALTER TABLE … DROP CONSTRAINT IF EXISTS statements to schema.sql so the legacy CHECK constraints are removed on every boot.
- Keep the columns as plain VARCHAR; rely on application-side enum parsing as the single source of truth.
- Confirm Hibernate ddl-auto=update does not re-create the constraints on subsequent boots.

## Deliverables

- Updated backend/src/main/resources/schema.sql with the two DROP CONSTRAINT IF EXISTS statements.

## Acceptance Criteria

- [ ] After deploy, `POST /api/reports` with `objectType: "WASHROOM"` (and any other new enum value) returns 201 Created.
- [ ] `SELECT conname FROM pg_constraint WHERE conrelid IN ('report_objects'::regclass, 'report_object_issues'::regclass) AND contype = 'c'` returns 0 rows on production.
- [ ] Existing reports continue to read/write correctly; no data loss.
- [ ] All tests pass.

## Deadline
09.05.2026

## Notes
The proper long-term fix is a versioned migration tool (Flyway or Liquibase) so each enum addition gets its own migration script. That is intentionally out of scope for this hotfix; without it, the same drift will recur the next time enum values change. Track as a separate issue.

alaaddinerenn:
# 📝 Description
Fixes #456 and partly fixes #151 

Supersedes #457 (the schema.sql-only fix in `fix/drop-legacy-enum-check-constraints`). The same constraint drop is now applied as a Flyway migration here, and merging both would stage the drop in two places.

Adds Flyway as a versioned database migration tool to handle DB tweaks Hibernate `ddl-auto=update` cannot express. Hibernate generated CHECK constraints on `report_objects.object_type` and `report_object_issues.issue_type` from the original 5 ObjectType / 24 IssueType enums; `ddl-auto` does NOT extend those constraints when new enum values are added later, so production inserts with WASHROOM, ROOM_SIGN, PEDESTRIAN_CROSSING, CURB_RAMP and the new issue types fail at the DB layer with a constraint violation that surfaces as 500 Internal Server Error.

`spring.flyway.baseline-on-migrate=true` with `baseline-version=1` lets Flyway adopt existing databases (created previously by `ddl-auto=update`) by treating them as if they were already at V1, then applying V2 onwards. `V1__init.sql` is intentionally empty — it serves only as the baseline marker so `flyway_schema_history` can be initialized on existing environments without trying to recreate tables that Hibernate already manages. `V2__drop_legacy_enum_constraints.sql` is the actual fix; it idempotently drops the two stale CHECK constraints with `IF EXISTS`, safe to run on environments that may or may not have them.

Future schema tweaks (constraint changes, manual ALTERs, indexes Hibernate cannot infer) should be written as `V<n>__<description>.sql` migration files under `db/migration/` instead of being appended to `schema.sql`. The long-term path is to move fully to Flyway with `ddl-auto=validate` and a baseline snapshot of the current schema, but that is intentionally out of scope for this hotfix; this PR only sets up the infrastructure and addresses the immediate constraint blocker.

# 📊 Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation

- [x] All tests passed

# 📸 Screenshots / Recordings
N/A

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/src/main/resources/db/migration/V2__drop_legacy_enum_constraints.sql
+++ b/backend/src/main/resources/db/migration/V2__drop_legacy_enum_constraints.sql
@@ -11,5 +11,5 @@
 -- IF EXISTS keeps this safe to apply on databases that may or may not
 -- have these constraints — e.g. fresh environments where the constraint
 -- never existed in the first place.
-ALTER TABLE report_objects        DROP CONSTRAINT IF EXISTS report_objects_object_type_check;
-ALTER TABLE report_object_issues  DROP CONSTRAINT IF EXISTS report_object_issues_issue_type_check;
+ALTER TABLE IF EXISTS report_objects        DROP CONSTRAINT IF EXISTS report_objects_object_type_check;
+ALTER TABLE IF EXISTS report_object_issues  DROP CONSTRAINT IF EXISTS report_object_issues_issue_type_check;
--- a/backend/src/main/resources/db/migration/V2__drop_legacy_enum_constraints.sql
+++ b/backend/src/main/resources/db/migration/V2__drop_legacy_enum_constraints.sql
@@ -1,0 +1,15 @@
+-- Drop legacy CHECK constraints on enum columns. Hibernate generated these
+-- from the original 5 ObjectType / 24 IssueType enums when ddl-auto=update
+-- first ran against this schema; ddl-auto does NOT extend existing CHECK
+-- constraints when new enum values are added later, so inserts using the
+-- newer values fail at the DB layer with a constraint violation.
+--
+-- The columns stay plain VARCHAR; enum validation is enforced application-side
+-- via @Enumerated(EnumType.STRING) and Spring deserialization, which already
+-- rejects unknown values at the JSON layer with a 400 before any DB write.
+--
+-- IF EXISTS keeps this safe to apply on databases that may or may not
+-- have these constraints — e.g. fresh environments where the constraint
+-- never existed in the first place.
+ALTER TABLE report_objects        DROP CONSTRAINT IF EXISTS report_objects_object_type_check;
+ALTER TABLE report_object_issues  DROP CONSTRAINT IF EXISTS report_object_issues_issue_type_check;
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -63,12 +63,18 @@
 			<scope>runtime</scope>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-flyway</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.flywaydb</groupId>
 			<artifactId>flyway-core</artifactId>
+			<version>12.4.0</version>
 		</dependency>
 		<dependency>
 			<groupId>org.flywaydb</groupId>
 			<artifactId>flyway-database-postgresql</artifactId>
+			<version>12.4.0</version>
 		</dependency>
 		<dependency>
 			<groupId>org.hibernate.orm</groupId>
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -63,6 +63,14 @@
 			<scope>runtime</scope>
 		</dependency>
 		<dependency>
+			<groupId>org.flywaydb</groupId>
+			<artifactId>flyway-core</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.flywaydb</groupId>
+			<artifactId>flyway-database-postgresql</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.hibernate.orm</groupId>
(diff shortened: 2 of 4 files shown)


backend/src/main/resources/db/migration/V2__drop_legacy_enum_constraints.sql
```diff
@@ -0,0 +1,15 @@
...
+--
+-- The columns stay plain VARCHAR; enum validation is enforced application-side
+-- via @Enumerated(EnumType.STRING) and Spring deserialization, which already
+-- rejects unknown values at the JSON layer with a 400 before any DB write.
+--
+-- IF EXISTS keeps this safe to apply on databases that may or may not
+-- have these constraints — e.g. fresh environments where the constraint
+-- never existed in the first place.
+ALTER TABLE report_objects        DROP CONSTRAINT IF EXISTS report_objects_object_type_check;
+ALTER TABLE report_object_issues  DROP CONSTRAINT IF EXISTS report_object_issues_issue_type_check;
```
efemantarci: On a fresh Postgres (empty DB, no Hibernate-created tables yet), Flyway runs V2 first and fails. This should solve it:

```suggestion
ALTER TABLE IF EXISTS report_objects        DROP CONSTRAINT IF EXISTS report_objects_object_type_check;
ALTER TABLE IF EXISTS report_object_issues  DROP CONSTRAINT IF EXISTS report_object_issues_issue_type_check;
```

alaaddinerenn: I made the changes. Could you review it again! @efemantarci

efemantarci (approved):
Nice!

Issue #459, opened 2026-05-09
[Web] Document VITE_MAPTILER_KEY in frontend/.env.example


efemantarci:
# 📝 Document VITE_MAPTILER_KEY in frontend/.env.example

## Description

`VITE_MAPTILER_KEY` is read by [frontend/src/pages/Home.jsx](frontend/src/pages/Home.jsx) for the dark-mode basemap but isn't listed in [frontend/.env.example](frontend/.env.example).

## Feature / Task Details

- Add `VITE_MAPTILER_KEY=` to `frontend/.env.example` with a short comment.

## Deliverables

- Updated `frontend/.env.example`.

## Acceptance Criteria

- [ ] `VITE_MAPTILER_KEY` is present in `frontend/.env.example` with a brief comment.

## Deadline
09/05/2026

## Notes

mustafa-5493:
Closed by PR #460.

PR #460, opened 2026-05-09
[Web] Pass MapTiler key to frontend build

mustafa-5493:
## 📝 Description
Adds `VITE_MAPTILER_KEY` to the frontend deploy job's build env so production tiles load from MapTiler instead of falling back to Stadia, which 401s outside localhost. Also enables `workflow_dispatch` so we can re-run the deploy manually after env changes.

## 📊 Type of Change
- [x] Bug fix
- [ ] New feature

## ⏱️ Estimated Review Time
- [x] < 5 min

### DIFF
--- a/.github/workflows/deploy-frontend.yml
+++ b/.github/workflows/deploy-frontend.yml
@@ -28,6 +28,7 @@ jobs:
         working-directory: frontend
         env:
           VITE_API_URL: ${{ secrets.VITE_API_URL }}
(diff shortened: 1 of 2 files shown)


efemantarci:
Nice. Can you also add it to the .env.example as well? We can solve #459 here too

efemantarci (approved):
Nice! Will handle it on the server

mustafa-5493:
Done. Added VITE_MAPTILER_KEY to frontend/.env.example.  Closes #459

PR #464, opened 2026-05-09
[Backend] Add Custom Routing Profiles With Up To 5 Named Bundles Per User

alaaddinerenn:
# 📝 Description

Related to #365

Lets registered users save up to 5 named bundles of `RoutingConstraint` values ("İş yolu", "Hafta sonu", etc.) alongside the built-in `RoutingPreset` enum. They can create, edit, delete, and activate these profiles; activation copies the constraints onto the user record so the existing avoid-polygon routing logic keeps working unchanged.

The wire shape of `/api/users/me/routing-preferences` stays backward-compatible — the new `activeCustomProfileId` and `customProfiles` response fields are additive, no existing field changes meaning.

**New endpoints under `/api/users/me/routing-profiles`:**
- `GET /` — list user's saved profiles
- `POST /` — create (returns 409 on duplicate name or after the 5-profile cap)
- `PUT /{id}` — patch name / constraints / travel mode
- `DELETE /{id}` — delete (resets the user's routing state to NONE if it was the active one)
- `POST /{id}/activate` — set the profile as active

**Other touches:**
- `RegisteredUser.activeCustomProfile` is a `@ManyToOne` so Hibernate emits the FK on `active_custom_profile_id`. No Flyway migration is needed — the hybrid Flyway/Hibernate setup runs Flyway before `ddl-auto`, so any migration referencing the new tables would no-op.
- `RoutingPreferencesController` and `CustomRoutingProfileController` annotated with `@SecurityRequirement(name = BEARER_SCHEME)` so Swagger UI's Authorize button actually flows the bearer token into requests.
- Switching to a built-in preset (or deleting the active custom profile) automatically clears `activeCustomProfile` to keep the UI consistent.
- Flyway disabled in the test profile (H2 PostgreSQL mode doesn't support some PG-only constructs we may add to migrations later).

# 📊 Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [ ] Documentation

- [x] All tests passed

# 📸 Screenshots / Recordings

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [ ] 5-15 min
- [x] > 15 min


backend/src/main/java/com/bounswe2026group1/backend/model/UserCustomRoutingProfile.java
```diff
@@ -0,0 +1,95 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @NotNull
+    @ManyToOne(fetch = FetchType.LAZY, optional = false)
+    @JoinColumn(name = "user_id", nullable = false)
+    @ToString.Exclude
+    @EqualsAndHashCode.Exclude
+    private RegisteredUser user;
```
efemantarci, line 52: Add `@OnDelete(action = OnDeleteAction.CASCADE)` so profile rows go away when their owner is deleted. Without this, the admin user-delete flow throws a FK violation.

```suggestion
    @NotNull
    @ManyToOne(fetch = FetchType.LAZY, optional = false)
    @JoinColumn(name = "user_id", nullable = false)
    @OnDelete(action = OnDeleteAction.CASCADE)
    @ToString.Exclude
    @EqualsAndHashCode.Exclude
    private RegisteredUser user;
```

Also add to the imports at the top of the file:

```java
import org.hibernate.annotations.OnDelete;
import org.hibernate.annotations.OnDeleteAction;
```

backend/src/main/java/com/bounswe2026group1/backend/model/RegisteredUser.java
```diff
@@ -91,6 +93,20 @@ public class RegisteredUser {
...
+     * active. Held as a {@code @ManyToOne} so Hibernate emits the foreign
+     * key automatically — the hybrid Flyway/Hibernate setup runs Flyway
+     * before {@code ddl-auto=update}, so referencing this column in a
+     * migration would not work.
+     */
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "active_custom_profile_id")
+    @ToString.Exclude
+    @EqualsAndHashCode.Exclude
+    private UserCustomRoutingProfile activeCustomProfile;
```
efemantarci, line 108: Pair with the cascade on the other side — `SET NULL` here breaks the FK cycle so deleting a user with an active custom profile actually succeeds (otherwise Postgres would refuse to delete the profile row that the user row still points at).

```suggestion
    @ManyToOne(fetch = FetchType.LAZY)
    @JoinColumn(name = "active_custom_profile_id")
    @OnDelete(action = OnDeleteAction.SET_NULL)
    @ToString.Exclude
    @EqualsAndHashCode.Exclude
    private UserCustomRoutingProfile activeCustomProfile;
```

Also add to the imports at the top of the file:

```java
import org.hibernate.annotations.OnDelete;
import org.hibernate.annotations.OnDeleteAction;
```

efemantarci (commented):
Nice job. There is only 1 small issue. Deletes should cascade

backend/src/main/java/com/bounswe2026group1/backend/model/RegisteredUser.java
alaaddinerenn: Done — added `@OnDelete(action = OnDeleteAction.CASCADE)` on `UserCustomRoutingProfile.user` and `@OnDelete(action = OnDeleteAction.SET_NULL)` on `RegisteredUser.activeCustomProfile`. Good catch on the FK cycle.

Heads-up for anyone testing locally: `ddl-auto=update` doesn't rewrite existing foreign keys, so a `docker compose -f docker-compose.local.yml down -v && up --build -d` is needed to pick up the new constraints. Prod is unaffected since these tables don't exist there yet — Hibernate will create them with the cascade rules baked in on first deploy.

efemantarci (approved):
Nice! Merging.

PR #490, opened 2026-05-09
[Web] Don't Flash Signup Password Rules Red on First Keystroke

### LINKED ISSUE #489: [Web] Signup password rules turn red on first keystroke instead of after blur
## 📝 Description
On the signup page, typing one character into the password field flips every still-unmet rule to red ("Password must…"). Discouraging and noisy while the user is mid-typing.

In `frontend/src/pages/Signup.jsx`, the visibility predicate is:

```js
const showFailure = (passwordTouched || password.length > 0) && !rule.passed
```

The `|| password.length > 0` short-circuit makes red appear before the user has finished. Should rely only on `passwordTouched` (blur or submit) so unmet rules stay grey while typing.

## Feature / Task Details
- **Steps to Reproduce:** Open `/signup`, type one character in the password field, observe four of five rules instantly turn red.
- **Expected:** While typing for the first time, unmet rules stay grey. Each rule turns green the moment it is satisfied. Unmet rules only turn red after the field loses focus or the user clicks Create Account.
- **Implementation:** Change the predicate in `Signup.jsx` to `const showFailure = passwordTouched && !rule.passed`. Verify the existing submit handler still sets `passwordTouched` so unmet rules turn red if the user tries to submit without blurring first.

## Deliverables
- One-line fix in `frontend/src/pages/Signup.jsx`.

## Acceptance Criteria
- [ ] Typing into the password field never turns rules red until the user blurs the field or clicks Submit.
- [ ] Each rule turns green the moment it is satisfied.
- [ ] All tests pass.

mustafa-5493:
Closes #489.

The signup password checklist used `(passwordTouched || password.length > 0)` to flip rules to red. The OR meant a single keystroke turned every still-unmet rule red ,discouraging while the user is mid-typing.

Tightened the predicate to `passwordTouched && !rule.passed`. Unmet rules now stay neutral grey while typing, turn green the moment they're satisfied, and only flip to red once the user blurs the field or hits Submit (the existing handler already sets `passwordTouched` on submit).

Updated the matching Vitest case so it asserts both phases (neutral during typing, red copy after blur).

## 📊 Type of Change
- [x] Bug fix

## ✅ Checklist
- [x] All tests passed (255/255)

## 🧪 How to Test
1. Open `/signup`. Type one character. None of the unmet rules turn red.
2. Keep typing. Rules turn green one by one as they're satisfied.
3. Tab away (or click Create Account with an invalid password). Remaining unmet rules now turn red with descriptive copy.

## ⏱️ Estimated Review Time
- [x] < 5 min

### DIFF
--- a/frontend/src/pages/Signup.jsx
+++ b/frontend/src/pages/Signup.jsx
@@ -151,8 +151,11 @@ function Signup() {
                     aria-live="polite"
                   >
                     {passwordValidation.results.map((rule) => {
-                      const showFailure =
-                        (passwordTouched || password.length > 0) && !rule.passed
+                      // Show red only after the user has tabbed away or
+                      // tried to submit (passwordTouched). While they're
+                      // still typing for the first time, unmet rules stay
+                      // grey so the form doesn't flash red on every keystroke.
+                      const showFailure = passwordTouched && !rule.passed
                       const colorClass = rule.passed
                         ? 'text-[#176a21]'
                         : showFailure
--- a/frontend/src/pages/Signup.test.jsx
+++ b/frontend/src/pages/Signup.test.jsx
@@ -158,11 +158,19 @@ describe('Signup page', () => {
   // ─── Password validation UI ──────────────────────────────────────────────────
 
   describe('password requirements checklist', () => {
-    it('shows a descriptive message for each unmet rule as the user types', async () => {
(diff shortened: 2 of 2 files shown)


delos0 (approved):
LGTM

PR #515, opened 2026-05-10
[Web] Persist agree/disagree button state across navigation

### LINKED ISSUE #514: [Web] Agree/disagree Vote Button Loses Selected State After Navigation
# 📝 [Web] Agree/disagree vote button loses selected state after navigation

## Description
On the web app, after a user clicks the **Agree** (or Disagree) button on a report, the button correctly turns green to indicate the active vote. However, if the user navigates to a different page via the sidebar and then returns to the same report, the button reverts to its default (unselected) appearance — even though the vote is still recorded on the backend. The visual state does not reflect the user's actual vote.

## Feature / Task Details
Two issues combine to produce this bug:

- The `Home` page keeps the user's per-report vote in a local React state (`userVotes`). When the user navigates away from `/`, `Home` unmounts and that local state is lost.
- `getReports()` and `getReportById()` in `reportService.js` are called without an `Authorization` header, so the backend cannot identify the caller and always returns `userVote: null` on each report. There is therefore no fallback when the local state is empty after re-mount.

Scope:
- `frontend/src/services/reportService.js` — forward the auth token to `GET /api/reports` and `GET /api/reports/{id}` so the backend can populate `userVote`.
- `frontend/src/hooks/useReports.js` — pull the token from `useAuth()` and pass it through.
- `frontend/src/pages/Home.jsx` — when the local `userVotes` map has no entry for the selected report, fall back to `selectedReport.userVote` from the cached/refetched data.

## Deliverables
- Updated `getReports` / `getReportById` signatures that optionally accept a token and send `Authorization: Bearer <token>`.
- Updated `useReports` / `useReport` hooks that forward the auth token.
- Updated `ReportPanel` prop wiring in `Home.jsx` that falls back to `selectedReport.userVote`.

## Acceptance Criteria
- [ ] Voting Agree (or Disagree) on a report turns the button green (or red) as before.
- [ ] After navigating to another page via the sidebar and returning to the same report, the button still reflects the user's vote.
- [ ] Reloading the page while logged in renders the correct vote state on every report the user has voted on.
- [ ] Anonymous users continue to see all buttons in the default unselected state.
- [ ] All tests pass.

## Deadline
10.05.2026

## Notes
Backend already supports per-caller `userVote` hydration on `GET /api/reports` and `GET /api/reports/{id}` when a Bearer token is supplied — no backend changes are required.

alaaddinerenn:
# 📝 Description
Fixes #514

The agree/disagree button on a report turned green when clicked, but reverted to the default unselected style after navigating away via the sidebar and returning to the same report. This PR persists the vote-button state across navigations.

Root cause was twofold:
- `Home.jsx` kept the user's vote in a local `userVotes` state, which was wiped when the route changed and `Home` unmounted.
- `getReports()` and `getReportById()` did not forward the auth token, so the backend always returned `userVote: null` and there was no source of truth to fall back to after re-mount.

Changes:
- `services/reportService.js` — `getReports(token)` / `getReportById(id, token)` now send `Authorization: Bearer <token>` when a token is provided.
- `hooks/useReports.js` — `useReports` / `useReport` pull the token from `useAuth()` and pass it through.
- `pages/Home.jsx` — `ReportPanel`'s `userVote` prop falls back to `selectedReport.userVote` when the local vote map has no entry.

# 📊 Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation

- [ ] All tests passed

# 📸 Screenshots / Recordings
N/A

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/frontend/src/hooks/useReports.js
+++ b/frontend/src/hooks/useReports.js
@@ -12,14 +12,19 @@ export const reportKeys = {
 /**
  * Fetch the full report list.
  * Falls back to polling every 30 s when SSE is disconnected.
+ *
+ * The auth token is forwarded so each report carries the caller's
+ * `userVote` — without it the agree/disagree buttons lose their selected
+ * state across navigations.
  */
 export function useReports() {
   const sseStatus = useSseStatus()
   const disconnected = sseStatus !== 'connected'
+  const { token } = useAuth()
 
   return useQuery({
     queryKey: reportKeys.lists(),
-    queryFn: () => getReports().then((data) => data.map(mapReport)),
+    queryFn: () => getReports(token).then((data) => data.map(mapReport)),
     staleTime: 30_000,
     refetchInterval: disconnected ? 30_000 : false,
   })
@@ -32,10 +37,11 @@ export function useReports() {
 export function useReport(id) {
   const sseStatus = useSseStatus()
   const disconnected = sseStatus !== 'connected'
+  const { token } = useAuth()
 
   return useQuery({
     queryKey: reportKeys.detail(id),
-    queryFn: () => getReportById(id).then(mapReport),
+    queryFn: () => getReportById(id, token).then(mapReport),
     enabled: id != null,
     staleTime: 30_000,
     refetchInterval: disconnected ? 30_000 : false,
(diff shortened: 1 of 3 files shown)


delos0 (approved):
Nice catch, thanks for fixing

PR #519, opened 2026-05-10
[Web] Sign-up password warnings only turn red on submit attempt (closes #518)

### LINKED ISSUE #518: [Web] Sign-up password warnings turn red when leaving the field with content
# 📝 [Web] Sign-up password warnings turn red when leaving the field with content

## Description
Follow-up to #489. The current behavior turns the password rules red on `blur`, regardless of how much was typed. Switching from the password input to the email input after typing a single character flashes every unmet rule red, premature feedback during normal form flow.

`Signup.jsx:143` sets `passwordTouched = true` on every blur. The comment at line 154 says the intent was "after the user has tabbed away or tried to submit," but in practice "tabbed away" fires the moment focus moves, even with one keystroke.

## Feature / Task Details
Replace the blur trigger with a submit trigger so unmet rules stay grey while the user is still working on the form, and only turn red after the user has actually attempted submission. Rules continue to go green in real time as they're satisfied.

## Deliverables
- Updated `frontend/src/pages/Signup.jsx`
- Updated `frontend/src/pages/Signup.test.jsx`

## Acceptance Criteria
- [ ] Typing one character in password and switching to email leaves password rules grey
- [ ] Rules go green in real time as they're satisfied
- [ ] After clicking Sign Up with an invalid password, unmet rules turn red
- [ ] After a successful submit attempt, the red persists until the rules are satisfied
- [ ] All existing tests pass; new test covers the field-switch case

mustafa-5493:
## 📝 Description
Closes #518. Follow-up to #489.

The previous fix turned unmet password rules red on `blur`. That fired on every field switch, typing one letter and tabbing to the email field flashed all rules red. Now the rules only turn red after the user actually clicks Create Account with an invalid password. While they're still working on the form, unmet rules stay grey; passed rules turn green in real time.

The Create Account button no longer disables itself when the password is invalid, it stays clickable so the user can submit, see the rules go red, and understand what's wrong. `handleSubmit` already short-circuits and sets `passwordTouched = true` for invalid passwords.

## 📊 Type of Change
- [x] Bug fix

## ✅ Acceptance Criteria
- [x] All tests passed (308/308)

Issue acceptance criteria:
- [x] Typing one character in password and switching to email leaves rules grey
- [x] Rules go green in real time as they're satisfied
- [x] Clicking Create Account with an invalid password turns unmet rules red
- [x] Red persists until the rules are satisfied

## 🧪 How to Test
1. Open `/signup`.
2. Type one letter in the password field, tab/click to the email field. Password rules stay grey.
3. Type a valid password, rules go green one by one.
4. Clear the password, type something weak (e.g. `abc`), tick terms, click Create Account. All unmet rules turn red.

## ⏱️ Estimated Review Time
- [x] < 5 min

### DIFF
--- a/frontend/src/pages/Signup.jsx
+++ b/frontend/src/pages/Signup.jsx
@@ -140,7 +140,6 @@ function Signup() {
                     autoComplete="new-password"
                     value={password}
                     onChange={(e) => setPassword(e.target.value)}
-                    onBlur={() => setPasswordTouched(true)}
                     aria-invalid={passwordTouched && !passwordValidation.isValid}
                     aria-describedby="password-requirements"
                     required
@@ -151,10 +150,10 @@ function Signup() {
                     aria-live="polite"
                   >
                     {passwordValidation.results.map((rule) => {
-                      // Show red only after the user has tabbed away or
-                      // tried to submit (passwordTouched). While they're
-                      // still typing for the first time, unmet rules stay
-                      // grey so the form doesn't flash red on every keystroke.
+                      // Show red only after the user has actually tried to
+                      // submit (passwordTouched). While they're still working
+                      // on the form, unmet rules stay grey — passed rules
+                      // turn green in real time.
                       const showFailure = passwordTouched && !rule.passed
                       const colorClass = rule.passed
                         ? 'text-[#176a21]'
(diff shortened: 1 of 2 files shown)


delos0 (approved):
Fine

PR #545, opened 2026-05-10
[Backend] Fix Routing UX: Constraints, Profiles, and ORS Preference

### LINKED ISSUE #544: [Backend] Fix No Preference Preset Avoiding All Obstacle Reports in Routing
# 📝 Fix No Preference Preset Avoiding All Obstacle Reports in Routing

## Description

The `RoutingPreset.NONE` ("No preferences") option currently behaves identically to the anonymous-user fallback: every VERIFIED obstacle report becomes an avoid polygon, so the Accessible Route alternative goes far out of its way to skirt obstacles the user has explicitly said they don't care about.

The current behaviour is documented in [`ObstacleService.buildAvoidPolygons`](backend/src/main/java/com/bounswe2026group1/backend/service/ObstacleService.java) ("When constraints is empty the baseline is preserved: every VERIFIED obstacle report becomes an avoid polygon — anonymous and NONE-preset users get the historical 'Accessible Route' behaviour"). That conflation between "anonymous" and "explicitly chose No Preferences" is the bug — they are semantically different inputs and should produce different routes.

## Feature / Task Details

- Distinguish two empty-constraint cases in `RouteController` and the obstacle-service surface:
  - **Anonymous caller** (no JWT, no profile) → keep current defensive behaviour: avoid every verified obstacle.
  - **Authenticated caller with `RoutingPreset.NONE`** → no avoidance at all; routes should behave like the unconstrained fastest route, with no obstacle polygons sent to ORS and no `hasObstacles` warnings.
- Pick a sentinel that `ObstacleService.buildAvoidPolygons(...)` and `findObstaclesOnPath(...)` interpret as "skip avoidance entirely". Suggested: pass `null` for the `NONE` case and keep `Set.of()` as the anonymous baseline. Update the methods + javadocs accordingly.
- Update [`RouteController.getRouteOptions`](backend/src/main/java/com/bounswe2026group1/backend/controller/RouteController.java) to pass that sentinel when `caller != null && caller.getPreferredPreset() == RoutingPreset.NONE`.
- Update or add tests covering: anonymous caller still avoids all, NONE-preset user avoids none, non-empty constraints still filter as today.

## Deliverables

- Updated `ObstacleService` with a clear contract for "skip avoidance" vs "baseline avoid all"
- Updated `RouteController` that passes the right sentinel based on caller identity + preset
- Refreshed unit/integration tests for `RouteService` + `ObstacleService`
- Updated javadocs / inline comments so the next reader understands the three-state contract

## Acceptance Criteria

- [ ] Authenticated user with `RoutingPreset.NONE` receives routes that do not avoid any verified obstacles (Accessible Route alternative either matches the Fastest route or is omitted)
- [ ] Anonymous user still receives the historical "avoid every verified obstacle" baseline
- [ ] Authenticated user with non-empty constraints still receives polygons filtered to the hazards they care about (no regression on the existing constraint-aware path)
- [ ] `hasObstacles` flag on a route response stays consistent with which polygons were actually avoided
- [ ] All tests pass
- [ ] Documentation updated (javadoc on `buildAvoidPolygons` / `findObstaclesOnPath`)

## Deadline

2026-05-14 (Final milestone)

## Notes

Pure backend change — no API contract or DTO change, so no follow-up work for mobile/web.

efemantarci:
# 📝 Fix Routing UX: Constraints, Profiles, and ORS Preference

Fixes #544

Four related fixes to the routing API surface, layered as four commits:

## 1. `fix: extend null-constraints opt-out to any empty constraint set`
Original scope of #544 — a signed-in user who explicitly opts out of obstacle avoidance shouldn't be treated as the anonymous baseline. Generalises the three-state contract: a `RoutingPreset.NONE` user, a CUSTOM-preset user with zero rules, and any other authed user whose stored constraint set is empty all produce `constraints = null` (skip avoidance entirely). Anonymous callers still pass `Set.of()` and get the defensive baseline.

## 2. `fix: use ORS shortest preference for pedestrian routing`
ORS's `recommended` preference biases foot-walking onto side streets in OSM-sparse areas. Concretely: on Nisbetiye Caddesi between Selçuklar and Akmerkez the route detoured 343 m through side streets when the direct sidewalk path is 208 m. `shortest` produces the natural human-intuitive route. Pinned by a new `OrsRoutingClientTest` case.

## 3. `feat: adapt routing alternative set to caller's preferred mode`
The old contract always returned three alternatives (Fastest + Accessible + Ramp-Assisted). Now:
- **Anonymous**: Fastest + walking Accessible + Wheelchair (3)
- **Authed non-wheelchair**: Fastest + walking Accessible (2)
- **Authed wheelchair**: Fastest + Accessible Route in WHEELCHAIR mode (2) — same wheelchair-routing + ramp-via-leg logic, just labeled "Accessible Route" because it IS that user's accessibility-aware alternative.

Anonymous callers go through the existing 1-arg overload; authed callers go through the 3-arg overload, which derives two booleans from `preferredMode` and dispatches to a private impl.

## 4. `refactor: rename "Ramp-Assisted Route" to "Wheelchair Route"`
`RouteService` internally picks the best of a direct wheelchair route and a multi-leg via mapped ramp — `"Ramp-Assisted Route"` was a misnomer when no nearby ramp existed. `"Wheelchair Route"` describes the alternative honestly. As a side effect this fixes a latent mismatch: the web `RoutePanel.jsx` polyline-color and the `Home.jsx` notice logic were both already keyed on `"Wheelchair Route"` and never fired against the old label.

# 📊 Type of Change
- [x] Bug fix (commits 1, 2)
- [x] New feature (commit 3 — preset-aware route set; minor response-shape change)
- [x] Refactor (commit 4 — label rename)

- [x] Project builds successfully (`./mvnw clean verify`)
- [x] Hard-to-understand bits are commented (three-state constraints contract, two-flag dispatch in `RouteService`)
- [x] Added/updated tests — 37 routing-area tests now pass:
  - `OrsRoutingClientTest` (8): added `fetchDirections_usesShortestPreference`
  - `RoutingPreferencesRouteIntegrationTest` (12): added 5 new tests covering NONE opt-out, CUSTOM-empty opt-out, anonymous 3-card set, wheelchair-mode Accessible, walking/no-preference drops Wheelchair
  - `RouteServiceTest`, `RouteControllerTest`, `ObstacleServiceTest` updated for label rename
- [x] Manual end-to-end check against the local docker stack: anonymous → 3 cards, authed walking → 2 cards, authed wheelchair → 2 cards with WHEELCHAIR-mode Accessible Route

# ⚠️ Minor API Contract Change
Response array length is now 1–3 (was always 3). Frontend already iterates over the array dynamically, so no breaking changes — verified that `Home.jsx` / `RoutePanel.jsx` render correctly for all three caller types.

# 📸 Screenshots / Recordings
N/A — backend changes verified via curl against running stack.

# ⏱️ Estimated Review Time
- [x] 15-30 min (four logical commits, but each is reviewable independently)

### DIFF
--- a/backend/src/main/java/com/bounswe2026group1/backend/service/RouteService.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/service/RouteService.java
@@ -41,7 +41,11 @@ public class RouteService {
      * alternative whose mode matches their preferred travel mode.
      *
      * @param request           start/end coordinates from the controller
-     * @param constraints       caller's selected routing constraints; empty for anonymous / NONE preset
+     * @param constraints       caller's selected routing constraints. {@code null}
+     *                          means the signed-in caller picked
+     *                          {@code RoutingPreset.NONE} and explicitly opted out
+     *                          of avoidance (issue #544); empty set means the
+     *                          anonymous baseline; non-empty filters by hazard.
      * @param preferredMode     caller's preferred travel mode; null for no preference
      */
     public List<RouteResponse> getRouteOptions(
@@ -52,8 +56,9 @@ public class RouteService {
         Location start = new Location(request.getStartLat(), request.getStartLon());
         Location end = new Location(request.getEndLat(), request.getEndLon());
 
-        // Build avoid polygons once — reused for routes 2 and 3.
-        // Empty constraints preserves the pre-#365 baseline (every VERIFIED obstacle becomes a polygon).
+        // Build avoid polygons once — reused for routes 2 and 3. Three-state
+        // contract on `constraints`: null skips avoidance entirely, empty set
+        // is the anonymous baseline (avoid all), non-empty filters by hazard.
         ObjectNode avoidPolygons = obstacleService.buildAvoidPolygons(constraints);
 
         List<RouteResponse> routes = new ArrayList<>();
--- a/backend/src/main/java/com/bounswe2026group1/backend/service/RouteService.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/service/RouteService.java
@@ -143,72 +143,71 @@ public class RouteService {
         RouteResponse bestWheelchair = null;
 
         if (includeWheelchairAlternative) {
+            // Candidate A: direct wheelchair route with obstacle avoidance
+            RoutingDirectionsResult wheelchairResult =
+                    fetchOrNull(start, end, TravelMode.WHEELCHAIR, avoidPolygons);
+            if (wheelchairResult != null) {
+                bestWheelchair = RouteResponse.builder()
+                        .routeLabel(wheelchairLabel)
+                        .distanceMeters(wheelchairResult.getDistanceMeters())
+                        .durationSeconds(wheelchairResult.getDurationSeconds())
+                        .mode(TravelMode.WHEELCHAIR)
+                        .geometry(wheelchairResult.getGeometry())
+                        .steps(wheelchairResult.getSteps())
+                        .hasObstacles(false)
+                        .build();
+            }
 
-        // Candidate A: direct wheelchair route with obstacle avoidance
-        RoutingDirectionsResult wheelchairResult = fetchOrNull(start, end, TravelMode.WHEELCHAIR, avoidPolygons);
-        if (wheelchairResult != null) {
-            bestWheelchair = RouteResponse.builder()
-                    .routeLabel(wheelchairLabel)
-                    .distanceMeters(wheelchairResult.getDistanceMeters())
-                    .durationSeconds(wheelchairResult.getDurationSeconds())
-                    .mode(TravelMode.WHEELCHAIR)
-                    .geometry(wheelchairResult.getGeometry())
-                    .steps(wheelchairResult.getSteps())
-                    .hasObstacles(false)
-                    .build();
-        }
+            // Candidate B: multi-leg route through nearest ramp entry/exit
+            Report ramp = obstacleService.findClosestRampInBoundingBox(start, end);
+            if (ramp != null && ramp.getEntryPoint() != null && ramp.getExitPoint() != null) {
+                Location rampA = ramp.getEntryPoint();
+                Location rampB = ramp.getExitPoint();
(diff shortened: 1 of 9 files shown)


backend/src/main/java/com/bounswe2026group1/backend/service/RouteService.java
```diff
@@ -164,6 +208,8 @@ public List<RouteResponse> getRouteOptions(
             routes.add(bestWheelchair);
         }
 
+        } // end if (includeWheelchairAlternative)
```
alaaddinerenn: The new if (includeWheelchairAlternative) { ... } wrapper isn't re-indented and is closed with a } // end if (...) marker comment. ~65 lines of nested logic at the wrong indent level is hard to follow during review and during future edits.

Also, Edge case: a signed-in user with RoutingPreset.NONE AND preferredTravelMode = WHEELCHAIR. The controller passes constraints = null, so avoidPolygons = null. With includeWheelchairAlternative = true, this emits an "Accessible Route" in WHEELCHAIR mode with no avoidance polygons applied.

That's defensible ("user opted out"), but it isn't covered by a test and isn't called out in the javadoc. Please add either:

A test in [RoutingPreferencesRouteIntegrationTest](vscode-webview://0ag9cbbnn1uvp979070cs7fa9in5tp8a4mh3vb9o1m3ohcpn9rfv/backend/src/test/java/com/bounswe2026group1/backend/controller/RoutingPreferencesRouteIntegrationTest.java) for NONE preset + WHEELCHAIR mode asserting the emitted set and verify(obstacleService, ...).buildAvoidPolygons(isNull()), or
A one-line comment in the javadoc on the 3-arg getRouteOptions noting that wheelchair callers who opted out still receive a wheelchair-mode Accessible Route, but without avoidance.

alaaddinerenn (changes requested):
Could you address this small issues? @efemantarci

efemantarci:
Can you check again? I did the fix

alaaddinerenn (approved):
Everything is in order now. Merging!

PR #567, opened 2026-05-10
[Web] Make my-location button always pan and surface geolocation errors (#566)

### LINKED ISSUE #566: [Web] My-location button on map doesn't visibly do anything after the initial auto-locate
## 📝 Description
The third button in the zoom stack on the map page (my-location, after zoom-in and zoom-out) appears not to work.

Root cause is in `frontend/src/pages/Home.jsx`:

- On mount, `GeolocateOnLoad` already calls `map.locate({ setView: true, maxZoom: 16 })`, so the map is centered on the user at zoom 16.
- Clicking the button calls the same thing again. The browser returns the cached geolocation, Leaflet's `setView` lands on the same lat/lng/zoom, and there's no animation , looks broken.
- If geolocation permission is denied or the device can't fix a location, the click fails silently. No `locationerror` listener is attached, so the user gets zero feedback.

## 📊 Type
Bug

## ✅ Acceptance Criteria
- [ ] Clicking the my-location button always produces a visible pan back to the user's location, even when already centered there.
- [ ] If geolocation is unavailable or denied, the user sees a toast.

## 🧪 Test
- Open the map, wait for the initial auto-locate, pan away, click my-location → map flies back.
- Click my-location while already centered → still animates (zoom snaps to 16 if not already there).
- Block location permission, click my-location → error toast appears.

## ⏱️ Review Time
~2 min

mustafa-5493:
## 📝 Description
Closes #566.

`ZoomControls` now takes `userLocation` and `onLocationError`. If we already have the user's location (initial auto-locate has resolved), the click calls `map.flyTo(...)` directly -> guarantees a visible pan even when the cached position matches the current view. Otherwise it falls back to `map.locate(...)` with a one-shot `locationerror` listener so permission denials surface as a toast.

## 📊 Type
Bug fix

## ✅ Acceptance Criteria
- [x] Re-clicking the my-location button after the initial auto-locate always animates back to the user's position
- [x] Geolocation errors raise a toast instead of failing silently
- [x] No regressions in zoom-in / zoom-out

## 🧪 Test
- Wait for initial auto-locate, pan/zoom away, click my-location → smooth fly-back
- Click my-location while already centered → still animates (zoom snaps to 16)
- Deny location permission, click my-location → error toast

## ⏱️ Review Time
~2 min

### DIFF
--- a/frontend/src/pages/Home.jsx
+++ b/frontend/src/pages/Home.jsx
@@ -103,8 +103,22 @@ function makeMarkerIcon(status, objectType, reportType, selected = false) {
   })
 }
 
-function ZoomControls() {
+function ZoomControls({ userLocation, onLocationError }) {
   const map = useMap()
+  const handleLocate = () => {
+    if (userLocation) {
+      // Re-clicks after the initial auto-locate would otherwise hit the cached
+      // position and produce no visible pan. flyTo guarantees the animation.
+      map.flyTo([userLocation.lat, userLocation.lng], 16)
+      return
+    }
+    const onError = () => {
+      map.off('locationerror', onError)
+      onLocationError?.()
+    }
+    map.once('locationerror', onError)
+    map.locate({ setView: true, maxZoom: 16 })
+  }
   return (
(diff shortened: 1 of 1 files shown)


efemantarci (approved):
Nice! Merging

Issue #594, opened 2026-05-11
[Backend] Self-delete endpoint for the authenticated user (1.1.1.2.12)


mustafa-5493:
## 📝 Description
Implements requirement **1.1.1.2.12**. Adds a `DELETE /api/users/me` endpoint so a registered user can permanently delete their own account. Personal data is erased; their reports / comments / validation votes are kept and anonymized to preserve map integrity.

The anonymization recipe already exists in [AdminUserService.deleteUser](backend/src/main/java/com/bounswe2026group1/backend/service/AdminUserService.java:94) (name → "Deleted User", email → `deleted_{id}@deleted.invalid`, password → `$DELETED$`, status → BANNED). Extract it into a shared helper and call it from both admin-delete and the new self-delete path.

Scope:
- New endpoint `DELETE /api/users/me` returning 204 on success
- Resolves the caller from the JWT (no path id, no admin guard)
- Refactor: move the anonymization block into a shared private helper used by both admin- and self-delete
- Last-admin guard still applies , if the caller is the only ADMIN, return 409 with the same message
- 401 on unauthenticated requests

Out of scope:
- Hard deletion of the row (we keep the anonymized row to maintain FKs)
- Email confirmation flow for the deletion

## 📊 Type
New feature

## ✅ Acceptance Criteria
- [ ] `DELETE /api/users/me` returns 204 for an authenticated regular user; on next `GET /api/users/{id}` the row shows the anonymized values
- [ ] Caller's prior reports / comments / votes remain visible, attributed to the anonymized user
- [ ] Last ADMIN deleting themselves returns 409
- [ ] Unauthenticated request returns 401
- [ ] Anonymization helper covered by service-level unit tests; controller-level test covers happy path + last-admin guard + 401

## 🧪 How to Test
```
TOKEN=$(curl -s -X POST http://localhost:8080/auth/login \
  -H "Content-Type: application/json" \
  -d '{"email":"alice@test.com","password":"Test1234!"}' | jq -r .token)

# delete self
curl -s -o /dev/null -w "%{http_code}\n" -X DELETE \
  -H "Authorization: Bearer $TOKEN" \
  http://localhost:8080/api/users/me   # 204

# row is anonymized
curl -s http://localhost:8080/api/users/<that-id> | jq '{name,email}'
# { "name": "Deleted User", "email": "deleted_<id>@deleted.invalid" }

# previously submitted reports still visible
```

## ⏱️ Estimated Review Time
~5 min

PR #610, opened 2026-05-11
[Backend] Clear bio + avatarUrl on user anonymization (closes #609)

### LINKED ISSUE #609: [Backend] Clear bio and avatarUrl on user anonymization
## 📝 Description
Follow-up to #594. `RegisteredUserService.anonymizeUser` currently clears `name`, `email`, `password`, and `status` but leaves `bio` and `avatarUrl` untouched. Both can be PII , a free-text bio may name the user, and the avatar URL still points at their face photo in S3. Requirement **1.1.1.2.12** says "the system shall erase their personal data," so these slip through today.

Scope:
- In `anonymizeUser`: set `bio = null` and `avatarUrl = null`
- If `avatarUrl` was non-null, delete the underlying S3 object via `S3MediaService.deleteFile(...)` so the photo doesn't survive the anonymization. Wrap in `try/catch` , a missing/already-deleted object shouldn't fail the user-delete flow.
- Both self-delete (#594) and admin-delete inherit the fix automatically since they go through the same helper.

Out of scope:
- Mobile UI for self-delete (separate `[Mobile]` issue).
- Wiping derived data (points, badges, routing prefs) : these are not PII per the requirement and stay attached to the anonymized row.

## 📊 Type
Bug / polish

## ✅ Acceptance Criteria
- [ ] After self-delete or admin-delete, `bio` is null on the row
- [ ] After self-delete or admin-delete, `avatarUrl` is null on the row
- [ ] The S3 object backing the previous avatar is removed (when present)
- [ ] A missing/failed S3 deletion does not roll back the anonymization
- [ ] Existing self-delete and admin-delete tests still pass; one new assertion covers the cleared fields

## 🧪 How to Test

mustafa-5493:
## 📝 Description
Closes #609. Stacks on #594. `RegisteredUserService.anonymizeUser` now clears `bio` and `avatarUrl` alongside name/email/password/status, and deletes the underlying S3 object so the photo doesn't outlive the account.

S3 deletion is wrapped in `try/catch` , a missing or unreachable object logs a warning but does not roll back the anonymization. The user-facing fields are already cleared in the DB by that point, so failing the request would leave an inconsistent state.

Both self-delete (#594) and admin-delete inherit the fix automatically , they share the same helper.

## 📊 Type
Bug / polish

## ✅ Acceptance Criteria
- [x] `bio` is null after self-delete or admin-delete
- [x] `avatarUrl` is null after self-delete or admin-delete
- [x] S3 object is deleted via `S3MediaService.deleteFile(...)` when present
- [x] S3 failure logs a warning but doesn't fail the anonymization (verified by test)
- [x] No avatar → no S3 call (verified by test)
- [x] Existing tests still pass

## 🧪 How to Test
```bash
# Set a bio + upload an avatar, then self-delete
TOKEN=$(curl -s -X POST http://localhost:8080/auth/login \
  -H "Content-Type: application/json" \
  -d '{"email":"test@test.com","password":"Test1234!"}' | jq -r .token)

curl -s -X DELETE -H "Authorization: Bearer $TOKEN" \
  http://localhost:8080/api/users/me -o /dev/null -w "%{http_code}\n"   # 204

curl -s http://localhost:8080/api/users/<id> | jq '{name,email,bio,avatarUrl}'
# { "name":"Deleted User", "email":"deleted_<id>@deleted.invalid", "bio":null, "avatarUrl":null }
```

## ⏱️ Estimated Review Time
~3 min

### DIFF
--- a/backend/src/main/java/com/bounswe2026group1/backend/service/RegisteredUserService.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/service/RegisteredUserService.java
@@ -45,8 +45,10 @@ public class RegisteredUserService {
     private final UserBadgeRepository userBadgeRepository;
     private final PointEventRepository pointEventRepository;
     private final LeaderboardService leaderboardService;
+    private final S3MediaService s3MediaService;
     private final PasswordEncoder passwordEncoder;
     private final JwtUtil jwtUtil;
+    private static final org.slf4j.Logger log = org.slf4j.LoggerFactory.getLogger(RegisteredUserService.class);
 
     // Password strength regex pattern: Minimum 8 characters, at least one uppercase letter, one lowercase letter, one digit and one special character
     private static final String PASSWORD_PATTERN = "^(?=.*[0-9])(?=.*[a-z])(?=.*[A-Z])(?=.*[@#$%^&+=!]).{8,}$";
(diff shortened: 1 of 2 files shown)


efemantarci (approved):
Ok. This solves the issue I talked about.

PR #616, opened 2026-05-11
[Mobile] Live Leaderboard And Profile Refresh On Points Change

alaaddinerenn:
# 📝 Description

Mobile counterpart of the leaderboard / profile points refresh flow. Depends on the backend PR landing first, since the broadcast that triggers these listeners (`POINTS_CHANGED` on the public SSE stream) comes from `GamificationService.applyDelta` — without it, the listeners attached here will simply never fire.

Changes:
- **Model — extend `SseEvent` for `POINTS_CHANGED`.** Added nullable `userId` and `points` fields and a dedicated `toString` branch. `reportId` stays a non-nullable `int` (defaults to `0` for events without one) so existing `REPORT_*` callers don't need to handle a new nullable type.
- **Leaderboard screen — refresh on any points change.** `LeaderboardScreen` now subscribes to `SseService.events` in `initState` and re-runs `_load()` on every `POINTS_CHANGED` event. A 250 ms debounce coalesces bursts — a single report transition (e.g. PENDING → VERIFIED at 5 agrees) pays out multiple users in quick succession (author bonus + voter-alignment payouts) and would otherwise trigger one fetch per affected user.
- **Profile screen — refresh only on the caller's own changes.** `ProfileScreen` subscribes the same way but filters `event.userId == AuthService.userId` so other users' point movements don't churn the screen. The reload is also suppressed while the user is mid-edit, so an inbound event can't overwrite unsaved form state.

Both subscriptions are cancelled in `dispose`. App-lifecycle pause/resume of the SSE stream is already handled by `SseService` itself.

# 📊 Type of Change
- [x] New feature

- [x] All tests passed

Test added:
- `sse_event_test`: parser populates `userId` and `points` for a `POINTS_CHANGED` payload, the missing `reportId` falls back to `0`, and `toString` switches to the points-flavoured format.

The existing report-flavoured parser tests still pass unchanged, confirming the new optional fields don't perturb the existing payloads.

# 📸 Screenshots / Recordings
N/A

# ⏱️ Estimated Review Time
- [x] < 5 min

### DIFF
--- a/mobile/lib/models/sse_event.dart
+++ b/mobile/lib/models/sse_event.dart
@@ -1,7 +1,9 @@
 import 'dart:convert';
 
-/// Typed representation of the backend PublicSseEvent.
-/// eventType is one of: 'REPORT_CREATED', 'REPORT_UPDATED', 'REPORT_DELETED', 'MEDIA_ADDED'.
+/// Typed representation of the backend public SSE payloads.
+/// eventType is one of: 'REPORT_CREATED', 'REPORT_UPDATED', 'REPORT_DELETED',
+/// 'MEDIA_ADDED', 'POINTS_CHANGED'. POINTS_CHANGED carries [userId] and
+/// [points] instead of a reportId; reportId stays 0 for those events.
 class SseEvent {
   final String eventType;
   final int reportId;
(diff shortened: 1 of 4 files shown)


mertcat (changes requested):
When the points are changed, the page is refreshed, so it may cause non-stop refreshing. Other than this edge case, it is excellent work!

efemantarci (approved):
I checked it out too. Looks nice! Mert's commit handles it

mertcat (approved):
Now, it works well. Good job!

PR #626, opened 2026-05-11
[Backend] Fix Wheelchair Routing for Ramp Reports

### LINKED ISSUE #625: [Backend] Fix Ramp Usage for Routing
# 📝 Fix incorrect ramp entry selection causing massive wheelchair route detours

## Description
The wheelchair routing algorithm frequently fails and returns a massive detour (e.g., a 949m route instead of using a nearby ramp) on curved campus paths. The system currently uses straight-line (`haversineMeters`) distance to decide which end of a ramp to enter. On winding roads, the ramp's exit point might be geometrically closer to the starting location, causing the system to route the wheelchair into the stairs side. OpenRouteService then rejects the path and returns `null`, forcing the system to fall back to a long, unnecessary detour.

## Feature / Task Details
- Update the "Candidate B" multi-leg routing logic in `RouteService.java`.
- Instead of using haversine distance to pick the entry point, the algorithm must decode the `fastestResult` walking path and check the chronological order of the ramp's points.
- Implement a `findMinIndex` helper method to determine which ramp point appears first on the walking path; this point should be designated as the `rampEntry`.
- Refactor `ObstacleService` to find ramps lying directly on the path (`findRampOnPath`).
- Retain the existing bounding-box and haversine logic as a fallback mechanism in case the walking path is unavailable.

## Deliverables
- A new `findMinIndex` helper method in `RouteService.java`.
- Updated Candidate B logic that dynamically assigns `rampEntry` and `rampExit` based on path indexing.
- A new `findRampOnPath` method in `ObstacleService.java`.

## Acceptance Criteria
- [ ] Wheelchair routes successfully traverse ramps on curved paths without defaulting to massive detours.
- [ ] The system determines ramp entry/exit directions based on path sequence, avoiding "crashing" into the wrong side of the ramp.
- [ ] The bounding-box fallback executes correctly if the primary walking path is null.
- [ ] All tests pass

## Deadline
Sprint End

## Notes

SimsekMucahit:
# 📝 Description
Fixes #625
**Context & Problem:**
The wheelchair routing algorithm was failing on curved campus paths and returning massive detours. The root cause was that the algorithm used a straight-line (haversine) distance to determine the entry point of a ramp. On winding roads, the ramp's exit point could be geometrically closer to the starting location than its entry point, causing the system to attempt entering the ramp from the wrong direction (crashing into stairs) and making OpenRouteService return `null`.

**Solution:**
- Added a `findMinIndex` helper method in `RouteService.java`.
- Updated "Candidate B" logic to determine the ramp's entry and exit points based on their actual chronological order on the `walkingPath` (using polyline indices) rather than simple haversine distance.
- Added a fallback mechanism to use the old bounding-box search if the `fastestResult` path is completely unavailable.

# 📊 Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation

- [x] All tests passed

# 📸 Screenshots / Recordings
# ⏱️ Estimated Review Time
- [ ] < 5 min
- [x] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/src/main/java/com/bounswe2026group1/backend/service/ObstacleService.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/service/ObstacleService.java
@@ -198,6 +198,53 @@ public class ObstacleService {
     }
 
     // -------------------------------------------------------------------------
+    // Ramp routing: found all ramps
+    // -------------------------------------------------------------------------
+
+    /**
+     * Identifies if any verified ramp reports lie directly on or very close to the 
+     * given path. This is used to detect when a standard "fastest" route actually 
+     * utilizes a ramp, which helps in labeling and calculating accessibility metrics.
+     *
+     * @return The first verified ramp report found on the path, or null if none exist
+     */
+    public Report findRampOnPath(List<Location> pathPoints) {
+        if (pathPoints == null || pathPoints.size() < 2) {
+            return null;
+        }
+
+        // 1. Define a search area (bounding box) around the entire path padded by PATH_BUFFER_METERS to catch nearby features.
+        double bufferDeg = PATH_BUFFER_METERS / METERS_PER_DEGREE;
+        double minLat = pathPoints.stream().mapToDouble(Location::getLatitude).min().orElseThrow() - bufferDeg;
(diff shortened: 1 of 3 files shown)


efemantarci (approved):
Nice! It solves the duatepe issue.

Image

PR #646, opened 2026-05-12
[Web] Login Page Misreports Network And Server Errors As Invalid Credentials

### LINKED ISSUE #645: [Web] Login Page Misreports Network And Server Errors As Invalid Credentials
# 📝 Login Page Misreports Network and Server Errors as Invalid Credentials

## Description
The login page shows `Invalid email or password.` for every thrown error, not just `401`s. When the backend is cold-starting or a request fails for network/CORS/5xx reasons, the user sees a bogus credential error. Re-clicking Sign In a moment later succeeds because the backend is now warm — so users learn to double-click Sign In and conclude credentials are flaky.

Root cause is in [frontend/src/pages/Login.jsx](frontend/src/pages/Login.jsx) — the `catch` block masks every error with the same generic message, originally to avoid leaking which emails exist:

```js
} catch (err) {
  setError('Invalid email or password.')
}
```

`apiFetch` doesn't expose the HTTP status on the thrown `Error`, so the page can't tell `401` apart from a transient failure even if it wanted to.

## Feature / Task Details
- Attach the HTTP status to errors thrown by `apiFetch` so callers can branch on it.
- In Login, keep the generic message **only** for `401`s (preserves the no-enumeration property).
- For any other status / network error, surface the real error message so transient backend failures don't read as bad credentials.

## Deliverables
- `apiFetch` throws `Error` instances with a `.status` field set to `res.status`.
- Login's catch branches on `err.status === 401` vs. everything else.
- Existing Login tests updated to attach `status: 401` to their mocked rejections so they still hit the generic-message branch.

## Acceptance Criteria
- [ ] Submitting wrong credentials still shows `Invalid email or password.` (no enumeration leak).
- [ ] A transient network/5xx failure shows a real error message instead of pretending the credentials were wrong.
- [ ] All Login tests pass.

## Deadline
2026-05-14 (final milestone)

## Notes
Signup already passes `err.message` through, so it doesn't share this bug — left alone.

efemantarci:
# 📝 Login Page Misreports Network and Server Errors as Invalid Credentials

Fixes #645

Login's catch block masks every thrown error as `Invalid email or password.`, which makes transient backend failures (cold-start, 5xx, network blips) look like wrong credentials — users re-click Sign In and the second attempt succeeds against a now-warm backend, making login feel flaky. Branch on HTTP status instead: keep the generic message on 401 (no email enumeration), surface the real cause otherwise. To make this possible, `apiFetch` now attaches `.status` to the thrown `Error`.

# 📊 Type of Change
- [x] Bug fix

- [x] All tests passed

# ⏱️ Estimated Review Time
- [x] < 5 min

### DIFF
--- a/frontend/src/pages/Login.jsx
+++ b/frontend/src/pages/Login.jsx
@@ -40,9 +40,14 @@ function Login() {
       const from = location.state?.from?.pathname ?? '/'
       navigate(from, { replace: true })
     } catch (err) {
-      // Always show a generic message regardless of the actual error to avoid
-      // leaking whether the email exists in the system.
-      setError('Invalid email or password.')
+      // Generic message on 401 to avoid leaking whether the email exists;
+      // for network/5xx errors, surface the real cause so a transient backend
+      // failure isn't misread as bad credentials.
+      if (err?.status === 401) {
+        setError('Invalid email or password.')
+      } else {
+        setError(err?.message || 'Could not reach the server. Please try again.')
+      }
     } finally {
       setIsLoading(false)
     }
--- a/frontend/src/pages/Login.test.jsx
+++ b/frontend/src/pages/Login.test.jsx
@@ -183,7 +183,7 @@ describe('Login page', () => {
     })
 
     it('shows an error message on failed login', async () => {
-      authService.loginUser.mockRejectedValue(new Error('Unauthorized'))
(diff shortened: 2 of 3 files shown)


alaaddinerenn (approved):
Now we have better error descriptions. Approving!

PR #651, opened 2026-05-12
[Backend] Permit ASYNC/ERROR Dispatches In Security Filter Chain To Suppress Spurious SSE Disconnect Warnings

### LINKED ISSUE #650: [Backend] Permit ASYNC/ERROR Dispatches In Security Filter Chain To Suppress Spurious SSE Disconnect Warnings
# 📝 Permit ASYNC/ERROR Dispatches In Security Filter Chain To Suppress Spurious SSE Disconnect Warnings

## Description

Every time a client of the authenticated notification SSE stream (`GET /api/sse/notifications/subscribe`) disconnects, the backend logs a warning like:

```
Cannot render error page for request [null] as the response has already been committed.
```

This shows up roughly once per SSE client disconnect (≈1/min historically, ~1/10 min on the current container) and is pure log noise — it does not affect functionality, but it pollutes logs and masks real issues.

## Feature / Task Details

Root cause (verified against [SecurityConfig.java](../blob/main/backend/src/main/java/com/bounswe2026group1/backend/config/SecurityConfig.java), [NotificationSseController.java](../blob/main/backend/src/main/java/com/bounswe2026group1/backend/controller/NotificationSseController.java), [JwtAuthFilter.java](../blob/main/backend/src/main/java/com/bounswe2026group1/backend/config/JwtAuthFilter.java)):

1. Initial `REQUEST` dispatch: `JwtAuthFilter` authenticates the SSE handshake via the `sse_token` cookie, controller returns an `SseEmitter`, response is committed with `text/event-stream` headers (including `X-Accel-Buffering: no`).
2. Servlet thread returns to the pool, `SecurityContextHolder` is cleared.
3. When the client eventually disconnects, Tomcat performs an `ASYNC` dispatch back through the filter chain to finalize the response.
4. `JwtAuthFilter` extends `OncePerRequestFilter`, whose `shouldNotFilterAsyncDispatch()` returns `true` by default — so the JWT filter does not run on the async dispatch and `SecurityContext` stays empty.
5. Spring Security 6's `AuthorizationFilter` runs on **all** dispatch types by default (this is the breaking change from SS5, which only ran on `REQUEST`). With an empty context, `.anyRequest().authenticated()` rejects the dispatch.
6. `ExceptionTranslationFilter` tries to convert this to a 401, but the response was already committed in step 1 — hence the warning.

The recommended Spring Security 6 fix is to permit `ASYNC` and `ERROR` dispatches in the authorization rules so re-dispatches are not re-authorized (the initial `REQUEST` already enforced auth, which is the only dispatch a client can originate).

- Affects only logging; no behaviour change for the SSE stream itself
- Safe because ASYNC/ERROR dispatches are container-originated, not client-initiated
- Matches the official Spring Security migration guidance

## Deliverables

- Update `SecurityConfig.securityFilterChain` to permit `DispatcherType.ASYNC` and `DispatcherType.ERROR` before the existing matchers
- Backend continues to build and all existing tests pass

## Acceptance Criteria

- [ ] `.dispatcherTypeMatchers(DispatcherType.ASYNC, DispatcherType.ERROR).permitAll()` added to the `authorizeHttpRequests` block in `SecurityConfig`
- [ ] `./mvnw clean verify` passes locally
- [ ] No regression in auth behaviour for `/api/sse/notifications/subscribe` or any other endpoint
- [ ] All tests pass
- [ ] Spurious `Cannot render error page for request [null]` warnings stop appearing on SSE client disconnects

## Deadline

Before final milestone (2026-05-14).

## Notes

One-line fix; the analysis above is the rationale. The same change also suppresses any analogous warning that would occur on `ERROR` re-dispatches for already-committed responses.

efemantarci:
# 📝 Permit ASYNC/ERROR Dispatches In Security Filter Chain To Suppress Spurious SSE Disconnect Warnings

Fixes #650

Adds `.dispatcherTypeMatchers(DispatcherType.ASYNC, DispatcherType.ERROR).permitAll()` to the authorization rules so Spring Security 6's `AuthorizationFilter` only enforces auth on the initial `REQUEST` dispatch. SSE controllers commit the response early; on the container-driven async re-dispatch (triggered by client disconnect) the `SecurityContext` is empty, so the default filter chain rejected the already-committed response and logged `Cannot render error page for request [null] as the response has already been committed.` This change keeps auth enforcement intact (initial REQUEST is the only dispatch a client can originate) while silencing the noise.

# 📊 Type of Change
- [x]  Bug fix
- [ ]  New feature
- [ ]  Refactoring
- [ ]  Documentation

- [x] All tests passed

# 📸 Screenshots / Recordings
N/A — the only observable effect is the disappearance of the warning on SSE disconnects.

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/src/main/java/com/bounswe2026group1/backend/config/SecurityConfig.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/config/SecurityConfig.java
@@ -18,6 +18,7 @@ import org.springframework.web.cors.CorsConfiguration;
 import org.springframework.web.cors.CorsConfigurationSource;
 import org.springframework.web.cors.UrlBasedCorsConfigurationSource;
 
+import jakarta.servlet.DispatcherType;
 import java.util.List;
 
 @Slf4j
@@ -54,6 +55,12 @@ public class SecurityConfig {
                 .csrf(AbstractHttpConfigurer::disable)
                 .sessionManagement(session -> session.sessionCreationPolicy(SessionCreationPolicy.STATELESS))
                 .authorizeHttpRequests(auth -> auth
+                        // SSE controllers commit the response on the initial REQUEST
+                        // dispatch; on async/error re-dispatches the SecurityContext is
+                        // empty, so Spring Security 6's AuthorizationFilter (which runs
+                        // on every dispatch type by default) would otherwise reject the
+                        // already-committed response and log a spurious warning.
+                        .dispatcherTypeMatchers(DispatcherType.ASYNC, DispatcherType.ERROR).permitAll()
                         .requestMatchers(ROUTING_AND_AUTH_PUBLIC).permitAll()
                         .requestMatchers("/swagger-ui/**", "/swagger-ui.html", "/api-docs/**").permitAll()
                     .requestMatchers(org.springframework.http.HttpMethod.GET,


alaaddinerenn (approved):
I think this also improve the performance of the serror. Nice implementation!

PR #658, opened 2026-05-12
[Web] Show Reverse-Geocoded Place Name on Feed Report Cards

alaaddinerenn:
# 📝 Description
Report feed cards previously rendered raw `latitude, longitude` digits whenever a report had no `locationLabel` stored on the backend (older rows / external imports). This change resolves the coordinates to a short place name via Nominatim so users always see a human-readable location.

- Added a small `reverseGeocodeService` wrapper around Nominatim's `/reverse` endpoint plus a `useReverseGeocode` React Query hook (24h `staleTime` so the shared cache dedupes nearby cards and survives navigation).
- Exposed `locationLabel` on the mapped report shape so consumers can tell a real backend-provided label apart from the coordinate fallback.
- Extracted the inline feed card into `ReportFeedCard`, which only fires the reverse-geocode query when `locationLabel` is missing; falls back to "Resolving location…" while pending and "Location unavailable" on failure.

# 📊 Type of Change
- [x] New feature

- [x] All tests passed

# 📸 Screenshots / Recordings
image

# ⏱️ Estimated Review Time
- [x] 5-15 min

### DIFF
--- a/frontend/src/hooks/useReverseGeocode.js
+++ b/frontend/src/hooks/useReverseGeocode.js
@@ -1,0 +1,14 @@
+import { useQuery } from '@tanstack/react-query'
+import { reverseGeocode, reverseGeocodeKey } from '../services/reverseGeocodeService.js'
+
+export function useReverseGeocode(latitude, longitude, { enabled = true } = {}) {
+  const key = reverseGeocodeKey(latitude, longitude)
+  return useQuery({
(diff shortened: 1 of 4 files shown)


efemantarci (commented):
Sorry. But nominatim has 1 req/s rate limit for IP. So we cannot load many reports. This probably will be slow or uneffective

PR #661, opened 2026-05-12
[Backend] Fix Report Delete Failing When Subscriptions Exist

alaaddinerenn:
# 📝 Description
Fixes the 500 Internal Server Error returned by `DELETE /api/reports/{id}` for reports created after the `ReportSubscription` feature landed (2026-05-01).

Every report author is auto-subscribed via `subscriptionService.subscribeInternal(...)` at create time, but `Report` had no cascade for the `report_subscriptions` side. The orphan row's FK on `report_id` blocked the parent delete, producing a 500 on the owner-delete path (and the same would silently affect admin delete and the FIXED-report lifecycle scheduler).

The fix adds `@OneToMany(mappedBy = "report", cascade = CascadeType.ALL, orphanRemoval = true)` for `List<ReportSubscription> subscriptions` on `Report`, so Hibernate deletes the subscription rows in the same transaction before removing the report. No schema migration is required — the existing FK is untouched; only the order of DELETE statements changes.

# 📊 Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation

- [x] All tests passed

# ⏱️ Estimated Review Time
- [x] < 5 min
- [ ] 5-15 min
- [ ] > 15 min

### DIFF
--- a/backend/src/main/java/com/bounswe2026group1/backend/model/Report.java
+++ b/backend/src/main/java/com/bounswe2026group1/backend/model/Report.java
@@ -101,6 +101,9 @@ public class Report {
     @OneToMany(mappedBy = "report", cascade = CascadeType.ALL, orphanRemoval = true)
     private List<FixRequest> fixRequests = new ArrayList<>();
 
+    @OneToMany(mappedBy = "report", cascade = CascadeType.ALL, orphanRemoval = true)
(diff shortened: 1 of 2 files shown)


efemantarci (approved):
Nice! Merging

PR #664, opened 2026-05-12
[Backend][Frontend][Mobile] Scope Issue Filter per Object Type in Report Feed

alaaddinerenn:
# 📝 Description
The feed's object/issue filter previously combined `objectType` and `issueType` as two independent `EXISTS` clauses. Picking `RAMP` + `MISSING` matched any report that had *some* RAMP object AND *some* object with MISSING — even when the MISSING was on a door, not the ramp. The Advanced Filters panels (web and mobile) reflected that loose pairing: clicking MISSING under RAMP also highlighted MISSING under DOOR, and the mobile bottom sheet flattened its internal pair set down to a plain issue list on apply.

**Backend** — `ReportFeedQuery` gains an `objectIssue` parameter where each entry is a scoped pair (`OBJECT_TYPE:ISSUE_TYPE`). The `EXISTS` clause now requires the type and issue to live on the *same* object. Bad pairs (malformed string, unknown enum) return 400. The legacy `objectType` / `issueType` parameters stay on the backend for API consumers that only want one-sided filtering.

**Frontend (web)** — `Feed.jsx`'s issue selection moves from a flat issue-key array to `OBJECT:ISSUE` pair strings. Toggling MISSING under RAMP only highlights and filters that specific pair; the rest of the groups stay untouched. Removing an object type prunes only its own pairs.

**Mobile** — `_AdvancedFiltersSheet` was already tracking selections as `Set<(ObjectType, IssueType)>` internally, but `_apply()` flattened the set down to a plain `List<IssueType>` on its way out. The sheet now emits `OBJECT_TYPE:ISSUE_TYPE` strings, the screen carries them through state, and `ApiService.getReportFeed` ships them as repeated `objectIssue` query params (replacing the old `issueType` parameter on the mobile client).

**UX fix bundled in** — `keepPreviousData` already kept the previous results visible during a refetch on web, but once the new (potentially much shorter) data arrived the list collapsed and the browser ended up scrolled to the top. The page now captures `window.scrollY` whenever the feed's queryKey changes and restores it via `useLayoutEffect` once the new data settles.

# 📊 Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation

- [x] All tests passed

Manual checks:
- [ ] **Web** — Picking RAMP and DOOR object types, then clicking MISSING under RAMP, only highlights the RAMP chip
- [ ] **Web** — A report with separate RAMP (no MISSING) and DOOR (with MISSING) objects does **not** appear when filtering `RAMP:MISSING`
- [ ] **Web** — Removing an object type drops only its own pair selections
- [ ] **Web** — Scrolling the feed and changing a filter keeps the scroll position roughly stable (no jump to top)
- [ ] **Mobile** — Same per-group toggle behaviour in the advanced filters bottom sheet
- [ ] **Mobile** — Applying RAMP+MISSING returns only reports whose ramp object is MISSING

# 📸 Screenshots / Recordings
image

# ⏱️ Estimated Review Time
- [ ] < 5 min
- [ ] 5-15 min
- [x] > 15 min


mertcat (approved):
Nice work!

